PR #19, opened 2018-03-07
Return 404 and 410 if object doesn't exist or is deleted.

agger-magenta:
Se diskussionen her: https://redmine.magenta-aps.dk/issues/21256#note-18

### DIFF
--- a/oio_rest/oio_rest/custom_exceptions.py
+++ b/oio_rest/oio_rest/custom_exceptions.py
@@ -34,6 +34,10 @@ class BadRequestException(OIOFlaskException):
     status_code = 400
 
 
+class GoneException(OIOFlaskException):
+    status_code = 410
+
+
 class DBException(OIOFlaskException):
 
     def __init__(self, status_code, message, payload=None):
--- a/oio_rest/oio_rest/db.py
+++ b/oio_rest/oio_rest/db.py
@@ -24,7 +24,7 @@ from authentication import get_authenticated_user
 from auth.restrictions import Operation, get_restrictions
 from utils.build_registration import restriction_to_registration
 from custom_exceptions import NotFoundException, NotAllowedException
-from custom_exceptions import DBException, BadRequestException, GoneException
+from custom_exceptions import DBException, BadRequestException
 
 """
     Jinja2 Environment
@@ -457,7 +457,8 @@ def delete_object(class_name, registration, note, uuid):
                 "No {} with ID {} found.".format(class_name, uuid)
             )
         if already_deleted_msg in e.message:
-            raise GoneException("This object has already been deleted.")
+            # DELETE is idempotent, no problem here
+            return
         if e.pgcode[:2] == 'MO':
             status_code = int(e.pgcode[2:])
             raise DBException(status_code, e.message)
--- a/oio_rest/oio_rest/db.py
+++ b/oio_rest/oio_rest/db.py
@@ -24,7 +24,7 @@ from authentication import get_authenticated_user
 from auth.restrictions import Operation, get_restrictions
 from utils.build_registration import restriction_to_registration
 from custom_exceptions import NotFoundException, NotAllowedException
-from custom_exceptions import DBException, BadRequestException
+from custom_exceptions import DBException, BadRequestException, GoneException
 
 """
     Jinja2 Environment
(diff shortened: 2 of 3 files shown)


oio_rest/oio_rest/db.py
```diff
@@ -448,9 +448,16 @@ def delete_object(class_name, registration, note, uuid):
...
+            'Error updating {} with uuid [{}], '
+            'Invalid [livscyklus] transition to [Slettet]'
+        ).format(class_name.lower(), uuid)
         if e.message == not_found_msg:
-            raise NotFoundException(e.message)
+            raise NotFoundException(
+                "No {} with ID {} found.".format(class_name, uuid)
+            )
+        if already_deleted_msg in e.message:
+            raise GoneException("This object has already been deleted.")
```
danchr: I believe this is incorrect according to [RFC7231](https://tools.ietf.org/html/rfc7231#section-4.2.2):

> A request method is considered “idempotent” if the intended effect on the server of multiple identical requests with that method is the same as the effect for a single such request.  Of the request methods defined by this specification, PUT, _**DELETE**_, and safe request methods are idempotent.

In my opinion, we should just ignore any attempt to delete something already deleted, just as we should ignore any attempt to do a no-op right.

oio_rest/oio_rest/custom_exceptions.py
```diff
@@ -34,6 +34,10 @@ class BadRequestException(OIOFlaskException):
     status_code = 400
 
 
+class GoneException(OIOFlaskException):
```
danchr: I think we should get rid of `OIOFlaskException`; in this case, there's [werkzeug.exceptions.Gone](http://werkzeug.pocoo.org/docs/0.14/exceptions/#werkzeug.exceptions.Gone) which does exactly the same.

oio_rest/oio_rest/oio_rest.py
```diff
@@ -259,7 +263,17 @@ def get_object(cls, uuid):
...
+            raise NotFoundException(
+                "No {} with ID {} found in service {}".format(
+                    cls.__name__, uuid, cls.service_name
+                )
+            )
+        # Raise 410 Gone if object is deleted.
+        if object[0]["registreringer"][0][
+            "livscykluskode"
+        ] == db.Livscyklus.SLETTET.value:
+            raise GoneException("This object has been deleted.")
```
danchr: Likewise.

oio_rest/oio_rest/db.py
agger-magenta: OK. Does that mean that an attempt to delete an object which is not found should also return OK? Or do we wish to uphold the distinction between deleting something which was there, but was deleted, and something which was never there or has been physically deleted?

oio_rest/oio_rest/custom_exceptions.py
agger-magenta: Good point, but should that be done as a part of this ticket? Getting rid of OIOFlaskException has more ramifications - e.g., we'd also need to implement a new DBException.

danchr: Not necessarily, although it's worth noting that the `DBException` doesn't work in the majority of cases…

oio_rest/oio_rest/oio_rest.py
agger-magenta: Likewise this is wrong according to the RFC, or we should get rid of the Gone exception?

danchr: Sorry; I was actually thinking about the latter, but the former also applies.

oio_rest/oio_rest/custom_exceptions.py
agger-magenta: But that should maybe be submitted as a separate bug. Consolidating our exceptions (in this case, getting rid of them) is a good idea, should it be done as a part of this ticket? Well, if the reviewers think so, I suppose so. :-)

oio_rest/oio_rest/db.py
danchr: I'm not sure what the point is in distinguishing the two, though; if it's easy to do, and something makes sense, sure — otherwise I wouldn't bother 🙂

It seems to me that returning `202 Accepted` is the most correct, actually, given that we don't perform any proper deletion right away. What do you think?

oio_rest/oio_rest/custom_exceptions.py
danchr: I'm just complaining ;)

oio_rest/oio_rest/oio_rest.py
agger-magenta: OK, you're right:

https://tools.ietf.org/html/rfc7231#section-6.5.9
"The 410 (Gone) status code indicates that access to the target
   resource is no longer available at the origin server and that this
   condition is likely to be permanent.  If the origin server does not
  know, or has no facility to determine, whether or not the condition
   is permanent, the status code 404 (Not Found) ought to be used
   instead."

The conclusion is that we should _not_ distinguish between deleted and non-existing objects in our responses. Unless maybe in the text message.

oio_rest/oio_rest/db.py
agger-magenta: I think it's nice to return a `200 OK` if we actually successfully (logically) deleted the object.

agger-magenta: We've agreed to return `202 Accepted` on all succesful DELETE operations. If the object does not exist, we'll still return a 404.

oio_rest/oio_rest/custom_exceptions.py
agger-magenta: We've agreed not to get rid of `OIO_FLASK_EXCEPTION` at this time.

oio_rest/oio_rest/oio_rest.py
agger-magenta: We've agreed to keep the `410 Gone` on the grounds that it might actually be useful for the caller to know that this object was deleted.

PR #23, opened 2018-03-20
Implemented verify_args to verify query arguments to oio_rest endpoints

cmoesgaard:
Implemented validation of query argument to oio_rest endpoints. Using unknown arguments now result in a `400 Bad Request` detailing the unknown arguments, as opposed to the argument just being ignored, as before.

Endpoints only strictly accept arguments that make sense in the given context, e.g. calling an objects `create` with _any_ query argument will result in an error, as the given endpoint doesn't use any arguments as all. This is an attempt to remove ambiguity and misunderstanding when using the API.

### DIFF
--- a/oio_rest/oio_rest/db_helpers.py
+++ b/oio_rest/oio_rest/db_helpers.py
@@ -82,6 +82,54 @@ def get_relation_names(class_name):
     return _relation_names[class_name.lower()]
 
 
+_search_params = {}
+
+GENERAL_SEARCH_PARAMS = {
+    'brugerref',
+    'brugervendtnoegle',
+    'foersteresultat',
+    'livscykluskode',
+    'maximalantalresultater',
+    'notetekst',
+    'uuid',
+    'vilkaarligattr',
+    'vilkaarligrel',
+}
+
+TEMPORALITY_PARAMS = {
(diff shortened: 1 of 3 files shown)


danchr:
Did you try merging this with the `minimox` branch and running the `mora` testsuite?

cmoesgaard:
No, but I'll try that now.

cmoesgaard:
After merging with the `minimox` branch and running the `mora` testsuite, everything appears to be working correctly.

cmoesgaard:
Having it as a decorator was indeed my original plan. The class reference doesn't exist when the decorator code is called, however, which made it difficult to generate the list of valid search terms specific to the class.

danchr:
> Having it as a decorator was indeed my original plan. The class reference doesn't exist when the decorator code is called, however, which made it difficult to generate the list of valid search terms specific to the class.

Well, the usual pattern is to have the decorator generate a new function which wraps the original definition; this new wrapper would then have full access to the instance object — and hence its class. This is what we do in [`util.restrictargs()` in MO](https://github.com/magenta-aps/mora/blob/development/mora/util.py#L162) — or perhaps I misunderstood you?

PR #24, opened 2018-03-26
Feature/21656 virkningstid

cmoesgaard:
Implemented 'virkningstid' and 'registreringstid', making it possible to view an objects state at specific points in time.
We already had the functionality in place as the default behaviour if no virkningstid was specified, I simply generalized it a bit to work on arbitrary points in time, and used the same approach for registrations.

### DIFF
--- a/oio_rest/oio_rest/oio_rest.py
+++ b/oio_rest/oio_rest/oio_rest.py
@@ -43,23 +43,21 @@ def typed_get(d, field, default):
 
 
 def get_virkning_dates(args):
-    virkning_fra = args.get('virkningfra', None)
-    virkning_til = args.get('virkningtil', None)
-    virkningstid = args.get('virkningstid', None)
+    virkning_fra = args.get('virkningfra')
+    virkning_til = args.get('virkningtil')
+    virkningstid = args.get('virkningstid')
 
-    if virkningstid and (virkning_fra or virkning_til):
-        raise BadRequestException("'virkningfra'/'virkningtil' are not "
-                                  "supported parameters with "
-                                  "'virkningstid'")
-
-    if virkning_fra is None and virkning_til is None:
-        if virkningstid:
-            # Timespan has to be non-zero length of time, so we add one
-            # microsecond
-            dt = dateutil.parser.isoparse(virkningstid)
-            virkning_fra = dt
-            virkning_til = dt + datetime.timedelta(microseconds=1)
-        else:
+    if virkningstid:
+        if virkning_fra or virkning_til:
+            raise BadRequestException("'virkningfra'/'virkningtil' conflict "
+                                      "with 'virkningstid'")
+        # Timespan has to be non-zero length of time, so we add one
+        # microsecond
+        dt = dateutil.parser.isoparse(virkningstid)
+        virkning_fra = dt
+        virkning_til = dt + datetime.timedelta(microseconds=1)
+    else:
+        if virkning_fra is None and virkning_til is None:
             # TODO: Use the equivalent of TSTZRANGE(current_timestamp,
             # current_timestamp,'[]') if possible
             virkning_fra = datetime.datetime.now()
@@ -69,15 +67,14 @@ def get_virkning_dates(args):
 
 
 def get_registreret_dates(args):
-    registreret_fra = args.get('registreretfra', None)
-    registreret_til = args.get('registrerettil', None)
-    registreringstid = args.get('registreringstid', None)
+    registreret_fra = args.get('registreretfra')
+    registreret_til = args.get('registrerettil')
+    registreringstid = args.get('registreringstid')
 
     if registreringstid:
         if registreret_fra or registreret_til:
             raise BadRequestException("'registreretfra'/'registrerettil' "
-                                      "are not supported parameters with "
-                                      "'registreringstid'")
+                                      "conflict with 'registreringstid'")
         else:
             # Timespan has to be non-zero length of time, so we add one
             # microsecond
--- a/oio_rest/oio_rest/oio_rest.py
+++ b/oio_rest/oio_rest/oio_rest.py
@@ -42,6 +42,51 @@ def typed_get(d, field, default):
     return v
 
 
+def get_virkning_dates(args):
+    virkning_fra = args.get('virkningfra', None)
+    virkning_til = args.get('virkningtil', None)
+    virkningstid = args.get('virkningstid', None)
+
+    if virkningstid and (virkning_fra or virkning_til):
+        raise BadRequestException("'virkningfra'/'virkningtil' are not "
+                                  "supported parameters with "
+                                  "'virkningstid'")
+
+    if virkning_fra is None and virkning_til is None:
+        if virkningstid:
+            # Timespan has to be non-zero length of time, so we add one
+            # microsecond
+            dt = parser.parse(virkningstid)
+            virkning_fra = dt
+            virkning_til = dt + datetime.timedelta(microseconds=1)
+        else:
+            # TODO: Use the equivalent of TSTZRANGE(current_timestamp,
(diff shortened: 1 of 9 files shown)


oio_rest/oio_rest/oio_rest.py
```diff
@@ -167,6 +168,51 @@ def _get_args(cls, as_lists=False):
...
+            else:
+                # TODO: Use the equivalent of TSTZRANGE(current_timestamp,
+                # current_timestamp,'[]') if possible
+                virkning_fra = datetime.datetime.now()
+                virkning_til = virkning_fra + datetime.timedelta(
+                    microseconds=1)
+        return virkning_fra, virkning_til
+
+    @classmethod
+    def get_registreret_timespan(cls, args):
```
danchr: Likewise.

oio_rest/oio_rest/oio_rest.py
```diff
@@ -167,6 +168,51 @@ def _get_args(cls, as_lists=False):
                                     request.args.getlist(k))
                 for k in request.args.keys()}
 
+    @classmethod
+    def get_virkning_timespan(cls, args):
```
danchr: AFAICT this method does not actually refer to `cls` in its code. As a result, it should either be a `@staticmethod` or a regular function — preferably the latter.

Furthermore, `timespan` is slightly odd nomenclature; perhaps `dates` or something of that ilk?

oio_rest/tests/util.py
```diff
@@ -185,11 +200,8 @@ def assertRequestFails(self, path, code, message=None, **kwargs):
 
     def _perform_request(self, path, **kwargs):
         if 'json' in kwargs:
-            # "In the face of ambiguity, refuse the temptation to guess."
-            # ...so check that the arguments we override don't exist
-            assert kwargs.keys().isdisjoint({'method', 'data', 'headers'})
-
-            kwargs['method'] = 'POST'
+            if 'method' not in kwargs:
+                kwargs['method'] = "POST"
```
danchr: Use `setdefault()`.

oio_rest/tests/util.py
```diff
@@ -185,11 +200,8 @@ def assertRequestFails(self, path, code, message=None, **kwargs):
 
     def _perform_request(self, path, **kwargs):
         if 'json' in kwargs:
-            # "In the face of ambiguity, refuse the temptation to guess."
-            # ...so check that the arguments we override don't exist
-            assert kwargs.keys().isdisjoint({'method', 'data', 'headers'})
```
danchr: Please either keep the two remaining checks, or let all three use `setdefault()`.

oio_rest/tests/util.py
```diff
@@ -231,6 +243,36 @@ def sort_inner_lists(obj):
...
+        if uuid:
+            path = "{}/{}".format(path, uuid)
+            r = self._perform_request(path, method='PUT',
+                                      json=get_fixture(fixture_name))
+        else:
+            r = self._perform_request(path, method='POST',
+                                      json=get_fixture(fixture_name))
+        return r
+
+    def get_json_result(self, path, **params):
```
danchr: Please add a docstring — also, what does this method do? The name isn't terribly descriptive…

oio_rest/oio_rest/oio_rest.py
cmoesgaard: I'll make them regular functions. Most of the functions in the file are implemented on the class, despite some of them not referring to the class at all, so it was simply an attempt to keep with that weird pattern.

I'll fix the name as well.

oio_rest/tests/util.py
cmoesgaard: I'm terrible at naming things.
What it does, is performing a `GET` towards oio_rest, while handling the nitty-gritty of unpacking the lists-of-lists structure of the result - preferably I would like to have used the MO `lora.py` functionality for this, but porting that over seemed a bit much.

danchr: Perhaps overload `assertRegistrationsEqual`? Or create an `assertQueryRespose`?

oio_rest/tests/util.py
cmoesgaard: All three checks now use `setdefault()`

oio_rest/tests/util.py
```diff
@@ -243,35 +242,44 @@ def sort_inner_lists(obj):
             sort_inner_lists(expected),
             sort_inner_lists(actual))
 
-    def load_fixture(self, path, fixture_name, uuid=None):
-        """Load a fixture, i.e. a JSON file with the 'fixtures' directory,
-        into LoRA at the given path & UUID.
+    def assertQueryResponse(self, expected, path, **params):
```
danchr: The argument order is inconsistent with `assertRequestResponse()` which takes `path` as the first argument, and `expected` as the second one:

```python
     def assertRequestResponse(self, path, expected, message=None,
                               status_code=None, drop_keys=(), **kwargs):		                               status_code=None, drop_keys=(), **kwargs)
```

Above all else, they should be consistent.

oio_rest/oio_rest/oio_rest.py
```diff
@@ -41,6 +42,51 @@ def typed_get(d, field, default):
     return v
 
 
+def get_virkning_dates(args):
+    virkning_fra = args.get('virkningfra', None)
```
danchr: You can drop the `, None` part — that's the default anyway…

oio_rest/oio_rest/oio_rest.py
```diff
@@ -41,6 +42,51 @@ def typed_get(d, field, default):
...
+    virkning_til = args.get('virkningtil', None)
+    virkningstid = args.get('virkningstid', None)
+
+    if virkningstid and (virkning_fra or virkning_til):
+        raise BadRequestException("'virkningfra'/'virkningtil' are not "
```
danchr: This phrasing feels a bit awkward to me — if nothing else, the passive voice is less common in English than in Danish. You could perhaps shorten it by simple stating that they conflict?

oio_rest/tests/util.py
```diff
@@ -231,6 +242,45 @@ def sort_inner_lists(obj):
...
+                                 'registreringstid'}:
+            actual = registrations
+        else:
+            assert len(registrations) == 1
+            actual = registrations[0]
+
+        return self.assertRegistrationsEqual(expected, actual)
+
+    def load_fixture(self, path, fixture_name, uuid=None):
+        """Load a fixture, i.e. a JSON file with the 'fixtures' directory,
```
danchr: `s/with/in/`

oio_rest/tests/util.py
```diff
@@ -231,6 +242,45 @@ def sort_inner_lists(obj):
...
+        into LoRA at the given path & UUID.
+        """
+        if uuid:
+            path = "{}/{}".format(path, uuid)
+            r = self._perform_request(path, method='PUT',
+                                      json=get_fixture(fixture_name))
+        else:
+            r = self._perform_request(path, method='POST',
+                                      json=get_fixture(fixture_name))
+        return r
```
danchr: You should verify or assert that the request succeeds rather than leaving this up to the caller.

oio_rest/oio_rest/oio_rest.py
```diff
@@ -41,6 +42,51 @@ def typed_get(d, field, default):
...
+                                  "supported parameters with "
+                                  "'virkningstid'")
+
+    if virkning_fra is None and virkning_til is None:
+        if virkningstid:
```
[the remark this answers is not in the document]
cmoesgaard: The logic in the two cases is slightly different. Here we have to specifically handle the case where `virkning_fra` and `virkning_til` is `None` and set them to today's date. I'll try to tighten up the logic a bit, though, to make the two functions seem more similar.

PR #34, opened 2018-04-17
Feature/21477 empty relation list fix

cmoesgaard:


oio_rest/oio_rest/oio_rest.py
```diff
@@ -318,7 +318,8 @@ def gather_registration(cls, input):
         """Return a registration dict from the input dict."""
         attributes = typed_get(input, "attributter", {})
         states = typed_get(input, "tilstande", {})
-        relations = input.get("relationer", None)
+        # Ensure that we only add 'relations' if the value is truthy
+        relations = input.get("relationer", None) or None
```
danchr: This comment isn't terribly helpful, considering that it merely states exactly what the code does. You should probably mention that otherwise, an empty list would override all relations.

Is there a particular reason why we can't change the default to be `[]` and handle that further on? That'd allow us to use `typed_get` for extra checking.

(Also, get rid of the `, None` while at it.)

oio_rest/oio_rest/oio_rest.py
```diff
@@ -318,7 +318,7 @@ def gather_registration(cls, input):
         """Return a registration dict from the input dict."""
         attributes = typed_get(input, "attributter", {})
         states = typed_get(input, "tilstande", {})
-        relations = input.get("relationer", None)
+        relations = typed_get(input, "relationer", {})
```
agger-magenta: Altså, her er bare en kort lille kommentar, egentlig et spørgsmål: Kan du bruge to minutter på at forklare, hvorfor dette her løser problemet? Umiddelbart ser det ud til, at begge kodestumper genererer et NULL i  `relations_array.sql` herunder.

agger-magenta: Ah, klart! Det var virkelig en gotcha. :+1:

PR #35, opened 2018-04-20
#21653 SAML upgrade

danchr:
This series switches us from `python-saml` to `python3-saml` which — despite its name — works on Python 2.7 as well.

The main stumbling block in the regard was the lack of test coverage for authentication. I've addressed this by adding integration tests that verify that we can verify a SAML assertion from AD FS, and that we fail in the appropriate scenarios. I was greatly aided in this effort by `python3-saml` being a good Python citizen that works with `freezegun`.


oio_rest/tests/test_integration_class.py
```diff
@@ -0,0 +1,299 @@
...
+
+from . import util
+
+
+class Tests(util.TestCase):
+    def test_import(self):
+        objid = self.load_fixture('/klassifikation/klasse',
+                                  'klasse_opret.json')
+
+        expected = {
```
agger-magenta: Is it a good idea to hard code the expected result? As I see it, it looks like a potential maintenance headache. Why not read the expected result from the fixture and process the JSON to match what we expect to get from LoRa?

danchr:
FWIW I updated this PR by adding a few comments to the test, reordering it a bit for readability, and rebasing it onto `development`.

oio_rest/tests/test_saml2.py
```diff
@@ -9,12 +9,13 @@ class TestSAML2(TestCase):
     def test_check_validity_raises_on_no_attribute_statement(self):
         # Arrange
         assertion_xml = ''
-        mox_entity_id = ''
-        idp_entity_id = ''
+        mox_entity_id = 'blyf'
```
agger-magenta: But why?

danchr: “blyf” is my preferred choice for a metasyntantic variable, closely followed by “flaf”, “klaf” and “kaflibob” — I find them more cromulent than “foo”, “bar” and “baz”.

On a more serious note, `python3-saml` validates these values, rejecting the empty string. You even get some particularly obnoxious error messages! (We should probably add some tests for those, someday…)

oio_rest/tests/test_integration_class.py
danchr: Yeah, it's a bit of a pain to change cases like this, but on the other hand, we also get incidental test coverage by asserting that the content doesn't change willy-nilly — which it most certainly shouldn't.

oio_rest/tests/test_saml2.py
agger-magenta: OK.

cmoesgaard:
More of a general question: Why can't we use OneLogin's library as it is? It almost seems to me that we're trying to roll our own solution to something that probably has already been solved.

danchr:
> More of a general question: Why can't we use OneLogin's library as it is? It almost seems to me that we're trying to roll our own solution to something that probably has already been solved.

Because we're not actually doing a full SAML authentication workflow, with a login and so on. That'd require a login page, or something like [their demo webapp](https://github.com/onelogin/python3-saml/blob/master/demo-flask/index.py). Essentially, that's a separate issue, although it seems equally likely that we'd simply implement something using OAuth instead.

danchr:
@agger-magenta Du you have any further requests or comments?

agger-magenta:
No, it looks good.

PR #36, opened 2018-04-23
 db: fix database generation

danchr:
I accidentally broke database generation with my Python 3.x
preparations. I did this by opting for a layout for states, which no
longer allowed simple instantiation as a dictionary. Instead, we
switch to a list of two-tuples.

### DIFF
--- a/oio_rest/tests/test_db_helpers.py
+++ b/oio_rest/tests/test_db_helpers.py
@@ -297,16 +297,14 @@ class TestDBHelpers(TestCase):
         with patch('oio_rest.db_helpers.db_struct', new={
             'testclass1': {
                 'tilstande': [
-                    'testtilstand1',
-                    [
+                    ('testtilstand1', [
                         'value1',
                         'value2'
-                    ],
-                    'testtilstand2',
-                    [
+                    ]),
+                    ('testtilstand2', [
                         'value3',
                         'value4'
-                    ]
+                    ]),
                 ],
             },
         }):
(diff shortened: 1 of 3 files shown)


oio_rest/tests/test_db_helpers.py
```diff
@@ -297,16 +297,14 @@ def test_get_state_names(self):
...
                         'value2'
-                    ],
-                    'testtilstand2',
-                    [
+                    ]),
+                    ('testtilstand2', [
                         'value3',
                         'value4'
-                    ]
+                    ]),
```
cmoesgaard: Why tuples here, when the 'real' structure uses lists?

danchr: speechless

danchr:
> Looks fine, on first impulse I'm not too happy about the move from dictionaries to lists due to the ordering issue, since we have a similar situation for attributes, where the input dict values also aren't ordered by definition, but I'm not sure I know how else to fix it, so that first impulse is probably most bikeshed painting more than anything else. And given the move, all of this looks perfectly sensible.

I'm inclined to agree, but unless we're willing to switch to Python 3.6, there's not much else to do :( (`dict` ordering is [stable in Python 3.6 and later](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-compactdict), although it's not actually a formal part of the language. However, since this change was adopted from PyPy — the only other Python implementation we might ever care about — it should be fine.)

PR #37, opened 2018-04-24
feature/22167 JSON-validator

andreaskring:
JSON-validator for checking the incoming JSON when creating new LoRa-objects.

(the code for "Dokument" is still not complete since we may have to change the structure of the OpretDokument JSON due to an inconsistency in LoRa with "Information- og meddelsesmodeller for Dokument".


andreaskring:
This PR needs a little update due to the DB_STRUCTURE changes in PR #36 . It is a small change that needs to be done in the JSON validator - will fix this

andreaskring:
The issue should be fixed now

oio_rest/oio_rest/validate.py
```diff
@@ -0,0 +1,356 @@
...
+    )
+
+
+def _adapter_tilstande(obj):
+    """
+    The 'tilstande' parts of DB_STRUCTURE was changed (upstream) to use
+    lists in some cases instead of dictoinaries. This function converts
+    the 'tilstande' parts back to dictionaries whenever necessary.
+    :param obj: The LoRa object type, i.e. 'organisation', 'bruger',...
+    :return: The 'tilstande' from DB_STRUCTURE for the given object type.
```
danchr: IMO you shouldn't refer to the temporal aspect — just explain what we do currently, and possibly what we should fix in the future.

Speaking of which, I fixed in `development` so that we now use two-tuples, which works with the `dict` constructor.

oio_rest/oio_rest/validate.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+    tilstande = db.REAL_DB_STRUCTURE[obj]['tilstande']
+    if type(tilstande) == types.DictType:
```
danchr: Use `isinstance(tilstande, dict)` instead.

oio_rest/oio_rest/validate.py
```diff
@@ -0,0 +1,356 @@
...
+        properties['tilstandsvaerdiattr'] = _generate_schema_object(
+            {
+                'forventet': BOOLEAN,
+                'nominelvaerdi': STRING
+            },
+            ['forventet', 'nominelvaerdi']
+        )
+        properties.pop('uuid')
+        relation_schema['tilstandsvaerdi']['items']['required'].remove('uuid')
+    if obj == SAG:
```
danchr: `elif`

oio_rest/oio_rest/db_structure.py
```diff
@@ -119,7 +125,8 @@
 
     "organisation": {
         "attributter": {
-            "egenskaber": ["brugervendtnoegle", "organisationsnavn"]
+            "egenskaber": ["brugervendtnoegle", "organisationsnavn"],
+            "required_egenskaber": ["brugervendtnoegle"],
```
danchr: You should lift this to the level above, so that it's consistent with `attributter_type_override`, etc. Perhaps even reuse that, if possible?

interface_test/test_data/indsats_opret.json
```diff
@@ -44,7 +44,7 @@
 		"relationer": {
 			"indsatsaktoer": [
 			{
-                                "index": 1,
+				"indeks": 1,
```
danchr: Nice catch — this is why we want to validate input!

oio_rest/oio_rest/db_helpers.py
```diff
@@ -61,7 +62,8 @@ def get_attribute_names(class_name):
             # specifically, the two state types of 'aktivitet' can
             # trigger occasional errors
             _attribute_names[c] = sorted(
-                c + a for a in db_struct[c]['attributter']
+                c + a for a in db_struct[c]['attributter'] if
+                not a == 'required_egenskaber'
```
danchr: You shouldn't hardcode `egenskaber`, as we might want to add support for other attributes in the future. If anything, do prefix comparison.

oio_rest/oio_rest/oio_rest.py
```diff
@@ -193,6 +195,13 @@ def create_object(cls):
         if not input:
             return jsonify({'uuid': None}), 400
 
+        # Validate JSON input
+        try:
+            obj_type = validate.get_lora_object_type(input)
+            jsonschema.validate(input, validate.SCHEMA[obj_type])
```
danchr: Please consolidate these two into a single helper function.

oio_rest/oio_rest/oio_rest.py
```diff
@@ -193,6 +195,13 @@ def create_object(cls):
...
+        except jsonschema.exceptions.ValidationError as e:
+            return jsonify({'message': e.message}), 400
```
danchr: We might want to add a general exception handler for these — or simply have the validation function return a boolean value.

oio_rest/oio_rest/validate.py
```diff
@@ -0,0 +1,356 @@
...
+        for key in db_attributter['egenskaber']
+    }
+    egenskaber.update({'virkning': {'$ref': '#/definitions/virkning'}})
+
+    egenskaber = _handle_special_egenskaber(obj, egenskaber)
+
+    return _generate_schema_object(
+        {
+            egenskaber_name: _generate_schema_array(
+                _generate_schema_object(egenskaber, db_attributter[
```
danchr: This is a somewhat unusual place to break the line — after the comma is more customary.

oio_rest/oio_rest/validate.py
```diff
@@ -0,0 +1,356 @@
...
+    :param obj: The type of LoRa object, i.e. 'bruger', 'organisation' etc.
+    :return: Dictionary representing the 'tilstande' part of the JSON schema.
+    """
+
+    tilstande = _adapter_tilstande(obj)
+
+    properties = {}
+    required = []
+    for key in tilstande.keys():
+        tilstand_name = '{}{}'.format(obj, key)
```
danchr: What are the types of `obj` and `key`? If both are strings, you can use `+` instead — it's simpler and more readable.

oio_rest/tests/test_oio_rest.py
```diff
@@ -268,7 +292,7 @@ def test_create_object_raises_on_unknown_args(self):
         # Act
         with self.app.test_request_context(method='POST',
                                            query_string=params), \
-                self.assertRaises(BadRequestException):
+             self.assertRaises(BadRequestException):
```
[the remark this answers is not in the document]
andreaskring: These whitespace changes was made by the IDE, when it was auto-correcting the indentions in the code

oio_rest/oio_rest/validate.py
andreaskring: Ok - very cool. I was not not aware of that feature of the `dict` constructur. In that case we do not need the adapter function at all. This makes things much simpler :-)

oio_rest/oio_rest/validate.py
andreaskring: No need to any longer - I have removed the adapter function according to the comment above. But I will keep the `isinstance` function in mind

oio_rest/oio_rest/validate.py
andreaskring: Ok - I will fix this

oio_rest/oio_rest/db_structure.py
andreaskring: IMHO it would be ok to leave it as it is. I think the `required_egenskaber` should be placed next to `egenskaber` in the `attributes` section. @agger-magenta and @cmoesgaard - do you have any opinion regarding this?

About the reuse question - I don't think it is possible, but if I am missing something, we should discuss it further

interface_test/test_data/indsats_opret.json
andreaskring: Thanks :-)

oio_rest/oio_rest/oio_rest.py
andreaskring: Yes - good point. i will fix this

oio_rest/oio_rest/oio_rest.py
andreaskring: Ok - we could look into this. Maybe we can leave it the way it is for now and then implement it in the next iteration?

oio_rest/oio_rest/validate.py
andreaskring: This was done by the IDE automatically - I will fix it to look a little nicer

oio_rest/oio_rest/validate.py
andreaskring: Yes - good point... we will use `+` instead

oio_rest/oio_rest/validate.py
```diff
@@ -0,0 +1,344 @@
...
+        'required': required,
+        'additionalProperties': False
+    }
+    if kwargs:
+        schema_obj.update(kwargs)
+    return schema_obj
+
+
+def _handle_special_egenskaber(obj, egenskaber):
+    if obj == KLASSE:
```
agger-magenta: Dette her bider måske sig selv i halen eller indeholder en rekursiv "hønen og ægget"-problematik, men i princippet kunne alle disse typer udledes ved at læse `db_structure.py`. Det er på en måde den, der er master for, hvad der er rigtigt eller forkert.

Alle disse her type-overrides bliver dermed i en forstand gentagelser af, hvad der allerede står i `db_structure`-modulet. 

Indholdet af dette skulle måske omvendt også checkes, men det må kunne gøres på en en anden måde end ved at opbygge dets overrides parallelt og separat.

agger-magenta: Ligeom i de følgende funktioner, mener jeg.

oio_rest/oio_rest/db_structure.py
danchr: > ```python
>         "attributter_override": {
>             "egenskaber": {
>                 "brevdato": {"type": "date", "mandatory": True},
>                 "major": {"type": "int"},
>                 "minor": {"type": "int"},
>                 "offentlighedundtaget": {"type": "offentlighedundtagettype"}
>             }
> ```

That seems nice — is there any particular reason why we couldn't simply overload the original `egenskaber` to do this?

agger-magenta: Well, it would clutter classes with only default behaviour.

agger-magenta: Basically, that proposal is the same as my 1) above. It is definitely an optionm the question is whether we want to keep the notion of a default behaviour with a number of "overrides".

danchr: I mean overloading in a literal sense — take the existing behaviour as a simple default, and allow a `dict` containing combined information when needed. For example, you'd combine:

```python
        "attributter": {
            "egenskaber": ["brugervendtnoegle", "beskrivelse", "brevdato",
                           "kassationskode", "major", "minor",
                           "offentlighedundtaget", "titel", "dokumenttype"]
        },
        "attributter_type_override": {
            "egenskaber": {
                "brevdato": "date",
                "major": "int",
                "minor": "int",
                "offentlighedundtaget": "offentlighedundtagettype"
            }
        },
```

Into:

```python

        "attributter": {
            "egenskaber": {
                "brugervendtnoegle": {
                    "mandatory": True,
                },
                "beskrivelse": {},
                "brevdato": {
                    "type": "date",
                },
                "kassationskode": {},
                "major": {
                    "type": "int",
                },
                "minor": {
                    "type": "int",
                },
                "offentlighedundtaget": {},
                "titel": {},
                "dokumenttype": {},
            },
        },
```

andreaskring:
Thanks :-)

The changes @danchr suggested have also been implemented, so I think we should be ready for the merge.

interface_test/test_data/tilstand_opret.json
```diff
@@ -74,7 +73,7 @@
             ], 
             "tilstandsvaerdi": [
               {
-                "indeks": 3, 
+                "indeks": 2,
```
danchr: Erm, why? Do we validate the sequential nature indices now? (If so, neat!)

oio_rest/oio_rest/db_helpers.py
```diff
@@ -138,9 +140,9 @@ def get_valid_search_parameters(class_name):
         # Add 'Dokument'-specific parameters not present in db_struct
         if _search_params.get('dokument'):
             _search_params['dokument'].update([
-                'varianttekst', 'deltekst'] +
-                get_document_part_relation_names()
-            )
+                                                  'varianttekst', 'deltekst'] +
```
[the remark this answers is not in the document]
danchr: Likewise; non-standard indentation. Probably triggers a warning too, somewhere.

oio_rest/oio_rest/db_helpers.py
```diff
@@ -88,7 +90,7 @@ def get_relation_names(class_name):
         for c in db_struct:
             _relation_names[c] = [
                 a for a in db_struct[c]['relationer_nul_til_en'] +
-                [b for b in db_struct[c]['relationer_nul_til_mange']]
+                           [b for b in db_struct[c]['relationer_nul_til_mange']]
```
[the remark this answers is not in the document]
danchr: The indentation here remains weird.

oio_rest/oio_rest/db_structure.py
```diff
@@ -96,11 +121,12 @@
...
-                "konfigurationreference"]
+                "konfigurationreference"],
         },
-        "attributter_type_override": {
-            "egenskaber": {
-                "konfigurationreference": "text[]"
+        "attributter_metadata": {
+            'egenskaber': {
+                'brugervendtnoegle': {'mandatory': True},
+                'konfigurationreference': {'type': 'text[]'}
```
danchr: I like what we settled on here; it's nice and readable!

oio_rest/oio_rest/db_structure.py
```diff
@@ -176,13 +217,18 @@
...
+        "attributter_metadata": {
+            'egenskaber': {
+                'afleveret': {'type': 'boolean'},
+                'beskrivelse': {'mandatory': True},
+                'brugervendtnoegle': {'mandatory': True},
+                'kassationskode': {'mandatory': True},
+                'offentlighedundtaget': {'type': 'offentlighedundtagettype'},
+                'principiel': {'type': 'boolean'},
+                'sagsnummer': {'mandatory': True},
+                'titel': {'mandatory': True},
```
danchr: On the other hand, this is getting a bit repetitive, but I suppose we have the standard authors to blame for that 😉

Probably too late now, but I see you have an `all` field below. Perhaps we should've stuck `mandatory` into a separate list…

oio_rest/oio_rest/db_structure.py
```diff
@@ -203,22 +249,38 @@
         "relationer_nul_til_mange": [
             "andetarkiv", "andrebehandlere", "sekundaerpart", "andresager",
             "byggeri", "fredning", "journalpost"
-        ]
-
+        ],
+        'relationer_metadata': {
+            'all': {
```
danchr: Perhaps we should call this one `'*'` instead of `all`? Someone might misread all the uses `all` to be proper names of relations, after all…

oio_rest/oio_rest/db_structure.py
```diff
@@ -303,43 +389,31 @@
...
-
-REAL_DB_STRUCTURE["aktivitet"]["relationer_type_override"] = {
-    "aktoerattr": "aktoerattr"
-}
-
-REAL_DB_STRUCTURE["tilstand"]["relationer_type_override"] = {
-    "tilstandsvaerdiattr": "vaerdirelationattr"
-}
+REAL_DB_STRUCTURE['klasse']['attributter_metadata']['egenskaber'][
+    'soegeord'] = {'type': 'soegeord'}
```
danchr: The indentation here is quite weird, but that's the fault of the somewhat horrible assignment. (Just complaining, don't really have any ideas for improvement, sorry…)

oio_rest/oio_rest/oio_rest.py
```diff
@@ -4,6 +4,7 @@
 import datetime
 
 import dateutil
+import jsonschema
 from flask import jsonify, request
 from custom_exceptions import BadRequestException, NotFoundException
```
danchr: We ought to insert a line break after this line, to distinguish 3rd party packages from our own. (Not your change, but still…)

oio_rest/tests/test_integration_create_organisation.py
```diff
@@ -152,8 +152,6 @@ def test_invalid_note(self):
         self.org['note'] = ['Note cannot be e.g. a list']
         self._check_response_400()
 
-    @unittest.skip(
-        'We are allowed to leave out the bvn - this should not be the case')
```
danchr: 👍

oio_rest/tests/test_integration_create_organisation.py
```diff
@@ -614,3 +602,113 @@ def test_relationer_key_not_allowed_twice(self):
         self.org = json.loads(org)
 
         self._check_response_400()
+
+    def test_create_itsystem(self):
+        """
+        TODO: move this elsewhere!!
```
danchr: If you're going to have a `TODO`, please make it more helpful. Reading this, I have no idea why this shouldn't be here, or where I ought to put it.

I'm guessing you probably meant that it doesn't belong in the `orgenhed` test, which is true. Please move it to a separate file, or rename the module if needs be.

oio_rest/tests/test_integration_create_organisation.py
```diff
@@ -614,3 +602,113 @@ def test_relationer_key_not_allowed_twice(self):
...
+        itsystem['livscykluskode'] = 'Opstaaet'
+        self.assertQueryResponse(
+            '/organisation/itsystem',
+            itsystem,
+            uuid=r.json['uuid']
+        )
+
+    def test_create_tilstand(self):
+        """
+        TODO: move this elsewhere!!
```
danchr: Do.

oio_rest/tests/test_oio_rest.py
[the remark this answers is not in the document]
danchr: Well, `flake8` warns about it, so please fix it 🙂

oio_rest/oio_rest/db_structure.py
agger-magenta: Fine, we should document this somewhere but it makes sense.

andreaskring:
@agger-magenta Ok - thanks.
@danchr if you can ok it too, I should be able to merge

oio_rest/tests/test_integration_create_organisation.py
```diff
@@ -98,10 +48,10 @@ def test_no_note_valid_bvn_no_org_name_no_relations(self):
         del self.org['attributter']['organisationegenskaber'][0][
             'organisationsnavn']
 
-        r = self._post(self.org)
+        r = self.post(self.URL, self.org)
 
         # Check response
-        self._check_response_201(r)
+        self.check_response_201(r)
```
danchr: You should be able to reuse the existing infrastructure for these and other tests, rather than adding new. I believe `flask_testing` has methods to this effect, and regardless, the name of the method is inconsistent with current practice: it should be `self.assertXXX` rather than `self.check_XXX`.

oio_rest/tests/util.py
```diff
@@ -341,6 +342,57 @@ class TestCase(TestCaseMixin, flask_testing.TestCase):
     pass
 
 
+class TestCreateObject(TestCase):
+    UUID_REGEX = re.compile('[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-'
```
danchr: Personally, I'd just use the `uuid` module instead — it's simple, foolproof and readable. A string is a valid UUID whenever you can use it to instantiate a UUID object. Something like this:

```python
def assertUUID(self, s: str):
    try:
        uuid.UUID(s)
    except (TypeError, ValueError):
        self.fail('{!r} is not a uuid!'.format(s))
```

oio_rest/tests/util.py
```diff
@@ -341,6 +342,57 @@ class TestCase(TestCaseMixin, flask_testing.TestCase):
...
+        Make HTTP POST request to Lora url
+        :param payload: dictionary containing payload to LoRa
+        :param url: E.g. '/organisation/organisation'
+        :return: Response from LoRa
+        """
+        r = self.client.post(
+            url,
+            data=json.dumps(payload),
+            content_type='application/json'
+        )
```
danchr: Use `self._perform_request` instead, which already takes a `json` argument. In fact, this method belongs on the superclass — there's nothing particularly create-object-y about it.

oio_rest/tests/util.py
```diff
@@ -341,6 +342,57 @@ class TestCase(TestCaseMixin, flask_testing.TestCase):
...
+        return r
+
+    def check_response_201(self, response):
```
danchr: `flask_testing` already has `self.assert200` and `self.assertStatus` — please rename this one to `self.assert201` for consistency, and make it use `assertStatus`.

oio_rest/tests/util.py
```diff
@@ -341,6 +342,57 @@ class TestCase(TestCaseMixin, flask_testing.TestCase):
...
+        """
+        Verify that the response from LoRa is 201 and contains the correct
+        JSON.
+        :param response: Response from LoRa when creating a new object
+        """
+        self.assertEquals(201, response.status_code)
+        self.assertEquals(1, len(response.json))
+        self.assertTrue(self.UUID_REGEX.match(response.json['uuid']))
+
+    def check_response_400(self, url, obj):
```
danchr: This method is trivial, and inconsistent in naming. Please either inline it, or rename it — I'd say inline it.

oio_rest/tests/util.py
```diff
@@ -341,6 +342,57 @@ class TestCase(TestCaseMixin, flask_testing.TestCase):
     pass
 
 
+class TestCreateObject(TestCase):
```
danchr: Generally speaking, this class feels superfluous. It adds a few utility methods which might as well belong to the superclass, and support for two fixed effects, which might either belong to the superclass anyway, or just be copied to the few subclasses.

oio_rest/tests/test_integration_create_organisation.py
```diff
@@ -595,7 +533,7 @@ def test_invalid_relation_name_not_allowed(self):
         self.org['relationer'] = {
             'unknown': [self.reference]
         }
-        self._check_response_400()
+        self.check_response_400('organisation/organisation', self.org)
```
danchr: `self.assertRequestFails()`

oio_rest/tests/test_integration_create_organisation.py
```diff
@@ -613,4 +551,4 @@ def test_relationer_key_not_allowed_twice(self):
             self.org['relationer']) + '}'
         self.org = json.loads(org)
 
-        self._check_response_400()
+        self.check_response_400('organisation/organisation', self.org)
```
danchr: `self.assertRequestFails()`

oio_rest/tests/test_integration_create_organisation.py
```diff
@@ -583,7 +521,7 @@ def test_reference_in_relation_must_be_an_uuid(self):
                 }
             ]
         }
-        self._check_response_400()
+        self.check_response_400('organisation/organisation', self.org)
```
danchr: `self.assertRequestFails()`

andreaskring:
I have done the following:

- Moved the `TestCreateObject` class into its own file, so it will no longer affect `util.py` (it is correct that it only has a few helper methods for now, but it will be needed for several integration tests to come and further helper methods will be added then.
- The `check_response_201`-method have been renamed to `assert201` and it is now using the `uuid` module to check the format of the UUIDs
- The `TestCreateObject.post()` have been removed and the `perform_request` is being used instead.
- The `check_response_400` method have been removed and we use `assertRequestFails` instead.

oio_rest/tests/test_integration_create_helper.py
```diff
@@ -0,0 +1,52 @@
...
+            "from_included": True,
+            "to": "2030-01-01 12:00:00+01",
+            "to_included": False
+        }
+        self.reference = {
+            'uuid': '00000000-0000-0000-0000-000000000000',
+            'virkning': self.standard_virkning1
+        }
+
+    def assertUUID(self, s):
```
danchr: This belongs in `util.TestCase`.

oio_rest/tests/test_integration_create_helper.py
```diff
@@ -0,0 +1,52 @@
...
+        try:
+            uuid.UUID(s)
+        except (TypeError, ValueError):
+            self.fail('{!r} is not a uuid!'.format(s))
+
+    def assert201(self, response):
```
danchr: Likewise, move to `util.TestCase`.

oio_rest/tests/test_integration_create_helper.py
```diff
@@ -0,0 +1,52 @@
...
+        Verify that the response from LoRa is 201 and contains the correct
+        JSON.
+        :param response: Response from LoRa when creating a new object
+        """
+        self.assertEquals(201, response.status_code)
+        self.assertEquals(1, len(response.json))
+        self.assertUUID(response.json['uuid'])
+
+    # def check_response_400(self, url, obj):
+    #     self.assertRequestFails(url, 400, json=obj)
```
danchr: Please remove this rather than commenting it out.

.gitignore
```diff
@@ -47,4 +47,5 @@ coverage.xml
 .coverage
 .eggs
 .pytest_cache
-venv
+python-testenv
+
```
danchr: Are these intentional?

andreaskring: Maybe not :-) After the change to Python 3 - is the virtual environments not called `python-env` and `python-testenv`?

PR #38, opened 2018-05-01
Jenkins docs

danchr:
This includes the documentation in our Jenkins setup. It's also viewable there, although without CSS styling, as that's apparently insecure…

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,21 @@ pipeline {
   }
 
   stages {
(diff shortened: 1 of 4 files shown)


danchr:
See also:

http://lorajenkins.atlas.magenta.dk/job/LoRa/job/feature%252Fjenkins-docs/
http://lorajenkins.atlas.magenta.dk/job/LoRa/job/feature%252Fjenkins-docs/Docs/

PR #39, opened 2018-05-01
Feature/21653 py3k

cmoesgaard:
Now that the [MO test suite successfully passes on python3 LoRa](http://lorajenkins.atlas.magenta.dk/job/LoRa/job/feature%252F21379-mora-testsuite/), I feel pretty confident in the Python3 LoRa conversion.

The code for running the MO test suite automatically will follow in a future PR.

### DIFF
--- a/oio_rest/oio_rest/app.py
+++ b/oio_rest/oio_rest/app.py
@@ -145,7 +145,8 @@ def log_api_call(response):
 
 @app.errorhandler(DataError)
 def handle_db_error(error):
-    message, context = error.message.split('\n', 1)
+    message = error.diag.message_primary
+    context = error.diag.context or error.pgerror.split('\n', 1)[-1]
     return jsonify(message=message, context=context), 400
 
 
--- a/oio_rest/oio_rest/app.py
+++ b/oio_rest/oio_rest/app.py
@@ -78,7 +78,7 @@ def get_token():
             text = tokens.get_token(username, password)
         except Exception as e:
             traceback.print_exc()
-            raise AuthorizationFailedException(e.message)
+            raise AuthorizationFailedException(*e.args)
 
         return Response(text, mimetype='text/plain')
 
--- a/oio_rest/oio_rest/app.py
+++ b/oio_rest/oio_rest/app.py
@@ -2,7 +2,7 @@
 
 import os
 import datetime
-import urlparse
+import urllib.parse
 import traceback
 
 from flask import Flask, jsonify, redirect, request, url_for, Response
@@ -112,7 +112,7 @@ def page_not_found(e):
 
 def get_service_name():
     'Get the hierarchy of the present method call from the request URL'
-    u = urlparse.urlparse(request.url)
+    u = urllib.parse.urlparse(request.url)
     urlpath = u.path
     service_name = urlpath.split('/')[1].capitalize()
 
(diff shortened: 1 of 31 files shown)


oio_rest/oio_rest/app.py
```diff
@@ -145,7 +145,8 @@ def log_api_call(response):
 
 @app.errorhandler(DataError)
 def handle_db_error(error):
-    message, context = error.message.split('\n', 1)
+    message = error.diag.message_primary
+    context = error.diag.context or error.pgerror.split('\n', 1)[-1]
```
agger-magenta: Kan vi få en forklaring på den indeksering?

oio_rest/oio_rest/auth/saml2.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, assertion_xml, mox_entity_id, idp_entity_id, idp_url,
                 },
                 sp_validation_only=True,
             ),
-            base64.b64encode(self.response),
+            base64.b64encode(self.response.encode('ascii')),
```
agger-magenta: Er self.response altid en ASCII-streng? Man risikerer ikke at få en encoding-error?

oio_rest/tests/test_oio_rest.py
```diff
@@ -237,7 +237,7 @@ def test_create_object_with_input_returns_uuid_and_code_201(self, mock):
                                            content_type='application/json',
                                            method='POST'):
             result = self.testclass.create_object()
-            actual_data = json.loads(result[0].data)
+            actual_data = json.loads(result[0].get_data(True))
```
agger-magenta: Why this `get_data(True)` thing instead of just accessing the data member?

danchr: The [`data` member](http://flask.pocoo.org/docs/1.0/api/#flask.Response.data) “should not be used and will eventually get deprecated.”

The use of the `True` value is probably wrong, though — I meant to pass `as_text=True` but AFAICT that's not what the code actually does. IIRC this has to do with `bytes` vs. `str` issues.

oio_rest/oio_rest/app.py
danchr: Yes, this uses the `context`, if set, and otherwise falls back to the second line of the general textual description — which was the previous behaviour. Could probably use a comment.

oio_rest/oio_rest/auth/saml2.py
danchr: I haven't seen anything but ASCII responses — but considering that this is XML, you're right, and we should probably just do a `utf-8` decode instead.

agger-magenta:
I'll just test the installer before approving (so please don't merge yet).

danchr:
> `ImportError: No module named parse`

Indeed, this is what happens when you run the code in Python 2.7…

cmoesgaard:
I'll take a look, and fix the installer.

danchr:
> I'll take a look, and fix the installer.

Please note that the agents haven't been converted even though they run in the same virtual environment. (Sorry.)

agger-magenta:
So if the agents aren't upgraded, they'll need their own virtual 
environment (or, as a workaround, they shouldn't be installed till 
further notice).

agger-magenta:
I also believe that the Python agents under `python_agents` aren't actually installed in the new pipeline project.

I think deprecating them until we need them shouldn't be a problem. But we need the install script to work just as well as the pipeline/Ansible/SaltStack/whatever provisioning _for the OIO REST API_ in order  to make a release.

danchr:
> …so I'll try to coordinate with him how to resolve this whole thing.

Have you tried the simple/stupid approach: switch the environment to Python 3 and see if it works?

cmoesgaard:
I've rebased on dev to get the new installer, and the installer has been updated to work with Py3k.
The installer has been tested on a new container on both mine and @istherepie's machine.

PR #42, opened 2018-05-03
simplify invocation of apply-templates, move into oio_rest, use during tests

danchr:
This makes it so that apply-templates itself obtains the database
structure, contains the list of templates, and extracts all templates
in one invocation. Finally, we move patch application into it as well.

This has three advantages:

1) By avoiding instantiating Python multiple times, we get a
   significant speedup.
2) It makes expanding the templates is much easier, both for
   development and testing.
3) By generating the database during testing, we decrease the chance
   of broken database changes slipping through.


db/initdb.sh
```diff
@@ -52,65 +52,7 @@ psql -d $MOX_DB -U $MOX_DB_USER -f funcs/_amqp_functions.sql
...
-patch --fuzz=3 -i  ../patches/dbtyper-specific_tilstand.sql.diff
-patch --fuzz=3 -i  ../patches/as_list_tilstand.sql.diff
-patch --fuzz=3 -i  ../patches/_remove_nulls_in_array_tilstand.sql.diff
-patch --fuzz=3 -i  ../patches/as_create_or_import_tilstand.sql.diff
-patch --fuzz=3 -i  ../patches/as_update_tilstand.sql.diff
-patch --fuzz=3 -i  ../patches/json-cast-functions_tilstand.sql.diff
-patch --fuzz=3 -i ../patches/as_search_tilstand.sql.diff
-
-cd ..
+$PYTHON ./apply-templates.py
```
cmoesgaard: `../oio_rest/apply-templates.py`?

PR #44, opened 2018-05-03
Feature/18459 install script replacement

istherepie:
This replaces the legacy installer.

Please note that this is an early pull request, 
the installer works but needs finetuning. 
However the changes fix certain coupling issues in the external deployment process.

Next step is to merge with the python 3 changes (Install process will break and needs minor fixing)


istherepie:
Please note that further code cleanup has not yet been completed.

istherepie:
Changed the installation directory and shell scripts as requested.

oio_rest/oio_rest/__init__.py
```diff
@@ -1,3 +0,0 @@
-
-# OIO-REST Python Package.
-# Dependencies: Must have access to the corresponding PostgreSQL database.
```
danchr: This change, along with others, moves the entire contents of the `oio_rest` package to the top-level module namespace. That's a rather unusual choice. Polluting the top-level namespace increases the probably of namespace collisions with other modules, including those distributed with Python as well as whatever third-party dependencies we might have. To quote the Zen of Python:

> Namespaces are one honking great idea -- let's do more of those!

Now, consolidating our sources into separate directories is a good thing, and I like the much cleaner directory layout in your branch. This one sticks out, though, and I believe we should re-instate something similar to the `oio_rest` module — perhaps in a sibling directory to `lib` called `src`.

oio_rest/setup.cfg
```diff
@@ -13,6 +6,7 @@ addopts = --cov=oio_rest --cov-report=xml --cov-branch --junitxml=tests.xml
 ignore = N999
 select = E,D,F,W,C,T,N,H
 exclude =
-        oio_rest/settings.py
+        settings.py
         .eggs,
-        venv*
+        .python-test
```
danchr: What's the justification for prefixing the virtual environment name with a `"."`? There's nothing particular hidden or specific about it, is there?

oio_rest/setup.cfg
```diff
@@ -1,10 +1,3 @@
-[egg_info]
-tag_build = dev
-tag_svn_revision = true
-
-[aliases]
-test=pytest
-
```
danchr: As you may know, `setup.cfg` is an alternative to `setup.py` that does much of the same, but with static configuration instead of code. Considering that you've removed `setup.py`, we should introduce configuration to `setup.cfg` that makes `oio_rest` similarly installable, as well also include some metadata that describes the package.

python_agents/README.rst
```diff
@@ -22,4 +22,16 @@ The agents have now been de-coupled and do not depend on external settings and p
...
+
+Installation
+------------
+A installation strategy (formula) has been created,
+however as previously described the agents have unresolved dependencies.
+
+As such they are not automatically installed as part of the stack.
+
+:NOTE:
+    Once the dependencies have been resolve,
```
danchr: +d

oio_rest/setup.cfg
agger-magenta: Or reintroduce `setup.py`.

oio_rest/settings.py
```diff
@@ -0,0 +1,85 @@
+# -*- python -*-
+
+import os
+
+# Environ wrapper
+env = os.environ.get
```
danchr: This name of this method is slightly misleading: `env` suggests that it's an environment, which typically means that it's a `dict`. But, it's really a getter method! Both the OS and Python provide a `getenv` pretty much exactly what you're looking for here — if needs be, consider doing `from os import getenv` instead, or use `os.getenv()` directly.

oio_rest/log_client.py
```diff
@@ -2,6 +2,7 @@
 import json
 
 import settings
+from settings import AUDIT_LOG_FILE
```
danchr: You already have `settings` imported in this module, so you can just use `settings.AUDIT_LOG_FILE` down below. The justification for this is that by doing an explicit import of the name, you've copied a reference into the namespace of this module. This means that this module no longer picks up changes to the reference in `settings`.

Generally speaking, I dislike such imports and try to avoid them myself. For `settings`, though, it's particularly since we frequently _do_ need to override them in tests, and it's nice do so only once, in a consistent manner.

python_agents/README.rst
istherepie: fixed

oio_rest/settings.py
istherepie: fixed

oio_rest/setup.cfg
istherepie: Initial setup.cfg was removed, however reintroduced solely for the flake8 and pytest configuration. OIO_REST is not installed as a package anymore.

oio_rest/tests/test_authentication.py
```diff
@@ -4,11 +4,11 @@
...
 from mock import MagicMock, patch
 
-from oio_rest import authentication
-from oio_rest.custom_exceptions import (AuthorizationFailedException,
+import authentication
+from custom_exceptions import (AuthorizationFailedException,
                                         UnauthorizedException)
 
-from . import util
+from tests import util
```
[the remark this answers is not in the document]
istherepie: There is no elaborate reasoning behind these changes,
the changes were part of fixing the import namespaces after moving oio_rest application to the top level in order to make the test suite work. 
It was my understanding that the PEP8 convention is "that absolute imports are recommended".

oio_rest/setup.cfg
istherepie: No justification - it was a matter of choosing a naming scheme that would less likely collide with any other virtual environments on a given system.
What name for the (TEST) virtual environment would you prefer ?

oio_rest/log_client.py
istherepie: fixed

oio_rest/oio_rest/__init__.py
istherepie: Would you prefer src at the top level or within an oio_rest namespace ?
(top level)
/src
/lib

(contained)
/oio_rest/src
/oio_rest/lib

istherepie:
Please note that the installer output is somewhat lacking and will be fixed in the upcoming days.

db/db-templating/apply-template.py
```diff
@@ -4,8 +4,8 @@
 import jinja2
 import os
 import sys
-from oio_rest.db_structure import DATABASE_STRUCTURE
-from oio_rest.db_structure import DB_TEMPLATE_EXTRA_OPTIONS
+from oio_rest_lib.db_structure import DATABASE_STRUCTURE
```
agger-magenta: As for naming, I'd personally prefer `oio_common` to `oio_rest_lib`. 

I'm not entirely convinced it should go into a `lib`directory all by itself. I think a structure like
```
oio_common/oio_common
^                     ^
|  setup.py       | db_structure.py
```
would be cleaner - i.e., drop the "lib" and put the `oio_common` at the top level.

One reason for this is that there's likely to be more functionality common to clients and server, and `oio_rest_lib` or `oio_common` is a candidate to be moved to a separate repository at some point. Just as the Java and Python agents. Or, if we retain the `lib`, we might consider moving `oio_rest` there too.

db/db-templating/generate-sql-tbls-types-funcs-for-oiotype.sh
```diff
@@ -6,7 +6,7 @@ MOXDIR=$(cd $DIR/../..; pwd)
 PYTHON="$MOXDIR/python-env/bin/python"
 export PYTHONIOENCODING=UTF-8
 
-oiotypes=$($PYTHON -m oio_rest.db_helpers)
+oiotypes=$($PYTHON -m oio_rest_lib.db_structure)
```
agger-magenta: Now that we're at it (and this is more of a general comment for all of us, I don't think it's important to do something about now), calling the `db_structure` module directly as an executable is not really very pretty. 

We should instead make a function to be configured as an entry point in  `setup.py` so that we get an executable in `bin`, e.g. of the virtual environment.

oio_rest/setup.cfg
danchr: > OIO_REST is not installed as a package anymore.

Although it's fine to leave that out of the installer, please reinstate the _ability_ for it to be installed. As an example, we'd like to use that for running tests in MO.

install/install_java.sh
```diff
@@ -1,65 +0,0 @@
-#!/bin/bash
```
agger-magenta: When we get to installing the MOX agents, how are we supposed to install Java and dependencies? I mean, maybe not using this for the time being should be preferred to removing it? Installing Java and dependencies might also be interesting as an option for the installer if some day we need to work with the Java agents again.

installer/base/tasks/install_java_components.sls
```diff
@@ -0,0 +1,72 @@
+# Import grains created by the task runner
```
agger-magenta: This seems to answer my question above regarding the Java dependencies. 

I believe it may not be necessary to add a Java 8 repository for Xenial? And I don't think we have to support Trusty from this version on.

Anyway, will this configuration be available from the installation script called by developers when creating a new VM?

lib/common/setup.py
```diff
@@ -0,0 +1,21 @@
+#
+# Copyright (c) 2017, Magenta ApS
```
agger-magenta: 2018 :-)

oio_rest/tests/test_authentication.py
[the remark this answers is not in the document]
danchr: > It was my understanding that the PEP8 convention is "that absolute imports are recommended".

Ah, I wasn't aware of that!

oio_rest/db_helpers.py
```diff
@@ -8,7 +8,7 @@
 from psycopg2.extensions import register_adapter as psyco_register_adapter
 from contentstore import content_store
 
-from db_structure import REAL_DB_STRUCTURE as db_struct
+from oio_rest_lib.db_structure import REAL_DB_STRUCTURE as db_struct
```
agger-magenta: Note that Andreas has some changes coming where the `REAL_DB_STRUCT` will be removed.

oio_rest/tests/test_authentication.py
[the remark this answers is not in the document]
istherepie: How would you like me to proceed ?

oio_rest/oio_rest/__init__.py
istherepie: Do you mean:

(top leve)
/src/oio_rest

If this is the case, do we want to keep "Python Agents" and "Mox Rest Frontend" as part of the repository or should these be moved their own repositories ?

oio_rest/setup.cfg
istherepie: Did I understand correctly that the test suite should have it's own virtualenv ? 
If this is the case I will rename it to python-testenv as to not conflict with a working/running virtualenv.

oio_rest/oio_rest/__init__.py
agger-magenta: I believe that `oio_rest` should still be a Python package. 

As such, the previous structure (`oio_rest/oio_rest`) was actually both ideal and standard.  We _could_ introduce an `oio_rest/oio_rest/src`, as some vesions of `paster` generate such directories, but I don't think it's necessary.

I think the structure should be more like
```
/oio_rest/ 
   db/
  oio_rest/   # with all the source moved back
  tests/
  runtests.sh
  setup.py
```
where the `setup.py` should include all requirements for the package itself, with tests being maybe out of scope. 

Whether or not you want the installer to install `oio_rest` as a package is a manner of convenience but I think it's very convenient if it is still installable as a package, and I think the most convenient and standardized way of doing this is keeping the old structure. (sorry)

agger-magenta: (On the other hand, if we retain it as a package, it may be most convenient for you to actually install it as a package alongside the other Python packages - that is, at any rate, a common way of doing things.

I concur that we should _not_ keep installing things in production with `setup.py develop`. )

oio_rest/setup.cfg
agger-magenta: Whatever. `venv`or `test-venv` or `python-test` would all work fine IMO.

oio_rest/tests/test_authentication.py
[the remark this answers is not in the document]
agger-magenta: I believe Dan introduced the relative imports as part of the Py3K integration. In all cases, as already written I'd like the application to be moved away from the top level so it works as a package again, so that might also be a good argument for reintroducing the relative imports. Especially if they're needed for the Py3K changes (Dan?).

oio_rest/oio_rest/__init__.py
agger-magenta: `oio_rest/src/` would work for me, but it's non-standard.

The standard would be `oio_rest/oio_rest/src`. But that's too verbose for me.

We _might_ have a `src` directory at the top level, but in that case all code-containing subdirectories (agents, python_agents, oio_rest, etc.) should be moved there. It's not a priority for me, but I wouldn't be opposed to that either.

oio_rest/tests/test_authentication.py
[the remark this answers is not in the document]
agger-magenta: Hmm, PEP8 says: 
```
Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):

import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example

However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:

from . import sibling
from .sibling import example

Standard library code should avoid complex package layouts and always use absolute imports.

Implicit relative imports should never be used and have been removed in Python 3.

```

I'd say keep absolute imports where possible, use relative where necessary.

oio_rest/setup.cfg
danchr: Yeah, my main gripe was with the `.` — anything else is up to you 🙂

oio_rest/setup.cfg
istherepie: The db_structure.py file has been extracted into its own package (mox_common_lib) as part of this change in order to allow $ROOT/db/initdb.sh to execute "python -m oio_rest_lib/db_structure". 

When recreating setup.py, mox_common_lib becomes redundant.

Would you like me to restore db_structure as part of the oio_rest package in this context instead of being part of mox_common_lib ?

db/db-templating/generate-sql-tbls-types-funcs-for-oiotype.sh
danchr: Actually, we can just make an `__main__.py` module in the package, and then invoke it with `-m` — that's nice and self-contained. Furthermore, since `oio_common` or `oio_rest_lib` contains nothing more than `db_structure`, couldn't we simply name it `oio_data` and obtain the list with `python -m oio_data`?

install/install_java.sh
danchr: > I mean, maybe not using this for the time being should be preferred to removing it?

IMO deleting it is fine — if needs be, we can always resurrect it from the version control history. I suspect the future installer would use a rather different approach anyway…

danchr:
For some reason or other, GitHub won't allow me to reply to Carsten's comments directly, so I'll do it here:

> As such, the previous structure (oio_rest/oio_rest) was actually both ideal and standard. We could introduce an oio_rest/oio_rest/src, as some vesions of paster generate such directories, but I don't think it's necessary.

The main issue with the previous structure is that it tended to lead to cluttering of the directory contents, which might well confuse newcomers. Although not terribly common, we could put them in an `src` directory which will, eventually, contain the entire sourcebase for both OIO REST and any bundled agents.

As an example, we'd have

```
setup.(py|cfg)
Jenkinsfile
NEWS
LICENSE
…
src/oio_rest
src/oio_(data|rest_common|lib)
…
```

> I believe Dan introduced the relative imports as part of the Py3K integration. In all cases, as already written I'd like the application to be moved away from the top level so it works as a package again, so that might also be a good argument for reintroducing the relative imports. Especially if they're needed for the Py3K changes (Dan?).

Python 3 removed support for _implicit_ relative imports, since they were rather unreliable, and tended to break things horribly. Both absolute imports and relative imports continue to work, but unlike in Python 2, the latter _must_ be prefaced with a dot to distinguish them from absolute imports.

db/db-templating/generate-sql-tbls-types-funcs-for-oiotype.sh
istherepie: Do I understand correctly that db_structure should remain as a seperate package from oio_rest - even if we do revert to installing oio_rest as a package again ?

install/install_java.sh
istherepie: Actually the installer includes a strategy for installing both java 8, compiling the artifact and installing mox rest frontend, it may need minor tweaking but should essentially work.

oio_rest/oio_rest/__init__.py
agger-magenta: >As an example, we'd have
```
setup.(py|cfg)
Jenkinsfile
NEWS
LICENSE
…
src/oio_rest
src/oio_(data|rest_common|lib)
```

My problem with that is that I think the `oio_common` package or whatever we call it should be installable independently from the `oio_rest` package. If we worry that the top level is becoming too cluttered, I'd prefer

```
/src/ 
    oio_rest/
         - setup.py, tests etc
          oio_rest/  # source files
   oio_common/
        - setup.py etc
       oio_common/  # source files
  python_agents/
       - source files, as now
  agents/
       - Java agents, as now
```

oio_rest/setup.cfg
agger-magenta: I'll just answer here too just to keep it on the record - no, having a common library is a good idea.

(But I think we should omit the `_lib` from the name).

db/db-templating/generate-sql-tbls-types-funcs-for-oiotype.sh
agger-magenta: @istherepie, yes. The separate package solves a use case for client applications to know the DB structure without installing the entire OIO REST package - which might, however, be used by other applications such as currently MO. So there is a need for two packages.

Maybe we should move some of the authentication stuff to the lib package too, but we can do that at a later time if needed.

install/install_java.sh
agger-magenta: OK!

agger-magenta:
Good! Regarding the `oio_common` package, the layout should be:

```
oio_common/
    oio_common/
        db_structure.py
        ... <in the future, maybe more source files> ...
    setup.py
```

i.e. the `setup.py` should be in the outermost of the two `oio_common` directories, with the actual code in the inner. That's similar to the conventions in other Python-packages, e.g. here: https://github.com/magenta-aps/bibos_admin/tree/master/bibos_client

istherepie:
I found no way for an explicit requirement of oio_common, as a workaround I have added both oio_common and oio_rest to a requirement.txt file.

danchr:
> I found no way for an explicit requirement of oio_common, as a workaround I have added both oio_common and oio_rest to a requirement.txt file.

You can specify the relative path between the two with a `-e` line, as per the [docs](https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format) for pip.

istherepie:
This is my workaround as well, however I found no valid way to set it in setup.py

agger-magenta:
>This is my workaround as well, however I found no valid way to set it in setup.py

You could have the `-e` line here, e.g.: https://github.com/magenta-aps/mox/blob/feature/18459_install_script_replacement/oio_rest/requirements.txt

A valid alternative would (IMO) be to upload the `oio_common` package to PyPI.

danchr:
> A valid alternative would (IMO) be to upload the oio_common package to PyPI.

I'd say we should use a relative path as long as they're in the same repository.

oio_rest/setup.py
```diff
@@ -30,17 +18,34 @@
...
+        'enum34==1.0.4',
+        'itsdangerous==0.24',
+        'psycopg2==2.6',
+        'wsgiref==0.1.2',
+        'python-saml==2.1.3',
+        'pexpect==3.3',
+        'python-dateutil==2.6.0',
+        'egenix-mx-base==3.2.9',
+        'pika'
+    ],
```
danchr: AFAICT this reverts almost all of `setup.py` to its state in `master`, and lowers numerous requirements to that state. Please revert it to `development` instead, which should fix tests, and also make installation easier. Among other things, that moved the list of requirements to `requirements.txt`.

istherepie:
Fixed dependencies and merged fixed conflicts with current development branch.

PR #46, opened 2018-05-09
Feature/21379 mora testsuite

cmoesgaard:
This adds functionality for automatically running the MO test suite on Jenkins.
The PR includes a few minor fixes to bugs which popped up through running the tests.


Jenkinsfile
```diff
@@ -32,12 +32,38 @@ pipeline {
...
+        }
+      }
+    }
+
+    stage('Test MO') {
+      steps {
+        dir('mora') {
+          timeout(15) {
+            ansiColor('xterm') {
+              sh "venv/bin/python -m pytest --verbose --junitxml=tests.xml tests --junit-prefix=MO"
```
danchr: Does the MO run affect coverage for the `oio_rest` package?

oio_rest/oio_rest/db.py
```diff
@@ -782,7 +782,7 @@ def search_objects(class_name, uuid, registration,
 def get_life_cycle_code(class_name, uuid):
     n = datetime.now()
     n1 = n + timedelta(seconds=1)
-    regs = list_objects(class_name, [uuid], n, n1, n, n1)
+    regs = list_objects(class_name, [uuid], n, n1, None, None)
```
danchr: (Would be nice to have `list_objects()` infer application time as well, but that's not an issue of this PR…)

oio_rest/oio_rest/oio_rest.py
```diff
@@ -324,10 +324,15 @@ def gather_registration(cls, input):
         """Return a registration dict from the input dict."""
         attributes = typed_get(input, "attributter", {})
         states = typed_get(input, "tilstande", {})
+
         relations = typed_get(input, "relationer", {})
+        filtered_relations = {
```
danchr: This could probably do with a comment — to a casual reader/reviewer, it's not obvious why we need this.

oio_rest/oio_rest/db.py
agger-magenta: Why is this change made?

oio_rest/oio_rest/oio_rest.py
agger-magenta: +1! I'm also not sure I get it (it might be OK for values to be Falsey?

oio_rest/oio_rest/db.py
danchr: Some issues with registration time and freezetime IIRC. Although reading the code, the root cause might be the explicit import of `datatime.datetime` that caches the object.

oio_rest/oio_rest/oio_rest.py
cmoesgaard: I'll my best to explain. 
I ran into an issue with the SQL generated from `relations_array.sql` specifically when we had relations with Falsy values at the end of the list of relations.

The template generates an array with a row for each relation, separated by commas. The commas are added whenever the row being generated is not the last in the list of relations. Whenever a relation is Falsy, row generation is skipped for that relation.

However, when the last relation in the list is Falsy, we end up with a trailing comma in the array, which is incorrect SQL. Due to the ordering of dicts being slightly different between py2 and 3, some of the MO tests began failing sporadically, due to us actually having Falsy relation values in some of our payloads - which due to pure luck, worked fine before with py2 oio_rest.

As far as I've been able to determine from the code, relations being Falsy didn't have any special meaning anywhere, so I decided that we might as well just filter them out early.

agger-magenta: OK!

agger-magenta: I think we could approve this, but the Jenkins tests are still failing.

agger-magenta:
@cmoesgaard, @andreaskring siger, på raketsnakken:

> Testen, du nævner, fejler fordi, der i denne fil https://github.com/magenta-aps/mora/blob/development/tests/fixtures/create_bruger_fætterguf.json bliver sendt `null` med på felterne for `note` og `objekttype`, og JSON validatoren forventer strenge ved disse felter. Felterne `note` og `objekttype` er ikke obligatoriske, så problemet kan løses ved blot at fjerne disse felter i JSON.

Dette er imidlertid i MORa-repoet, så det skal ændres der, hvis det skal ændres. Alternativt kan testkørslen fra LoRa (i Jenkins eller hvordan det nu er) sættes op til at forvente, at netop denne test fejler?

PR #47, opened 2018-05-09
Fix urn json validator

andreaskring:
Fix for JSON validator: URNs are now also accepted af references for relations


oio_rest/tests/test_validate.py
```diff
@@ -830,15 +937,19 @@ def test_valid_equivalence_classes3(self):
 
     def test_valid_equivalence_classes4(self):
         """
-        Equivalence classes covered: [71][74][76]
+        Equivalence classes covered: [71][74][76][112]
         See https://github.com/magenta-aps/mox/doc/Systematic_testing.rst for
         further details
         """
+        urn = {
+            'urn': 'This is an URN',
```
agger-magenta: A URN must start with "urn:". I.e., for something to be a valid URN, it should be: 'urn:This is an URN'.

And `oio_rest` actually checks for that. That is, this check should fail.

agger-magenta: Basically, a URN is any string that starts with "urn:" or "URN:". There is a lot about this in the corresponding RFC (https://tools.ietf.org/html/rfc8141), but LoRa is somewhat pragmatic and accepts all string starting with "urn:" or "URN:" as URNs. But that _is_ a condition. I don't know if there's an easy way of catching this in the JSON schema.

PR #50, opened 2018-05-15
#22658 — minor fixes & optimisations

danchr:
This series of commits contains changes that ease running LoRA — and
OIO REST in particular — in development mode. In particular, it allows
running `initdb.sh` without specifying any environment variables and
without restarting Gunicorn.

With these changes, I can do:

```ShellSession
$ python3.5 -m venv python-env
$ ./python-env/bin/pip install -q -e oio_rest -r oio_rest/requirements-test.txt gunicorn
$ dropdb --if-exists mox
$ dropuser --if-exists mox
$ db/initdb.sh
Password:
ALTER ROLE
GRANT
CREATE EXTENSION
CREATE EXTENSION
CREATE EXTENSION
CREATE EXTENSION
CREATE EXTENSION
GRANT
WARNING:  amqp can't find broker 1
 exchange_declare 
------------------
 f
(1 row)

CREATE SCHEMA
ALTER DATABASE
ALTER DATABASE
ALTER DATABASE
[snip]
$ LOG_AMQP_SERVER='' AUDIT_LOG_FILE='' python-env/bin/gunicorn \
> --pythonpath=oio_rest --threads=4 --reload --access-logfile - oio_rest.app:app
[2018-05-15 17:26:48 +0200] [67069] [INFO] Starting gunicorn 19.8.1
[2018-05-15 17:26:48 +0200] [67069] [INFO] Listening at: http://127.0.0.1:8000 (67069)
[2018-05-15 17:26:48 +0200] [67069] [INFO] Using worker: threads
[2018-05-15 17:26:48 +0200] [67072] [INFO] Booting worker with pid: 67072
```

In addition, it moves the support for running a LoRA server within the test suite to the `oio_rest` package so that we can easily re-use it from `mora` without relying on LoRA internals or duplicating the code.

### DIFF
--- a/db/initdb.sh
+++ b/db/initdb.sh
@@ -4,6 +4,16 @@ set -b
 MOXDIR=${BASE_DIR}
 DIR=${DB_DIR}
 
+test -z "$DIR" && DIR=$(cd $(dirname $0); pwd)
+test -z "$BASE_DIR" && BASE_DIR=$(cd $(dirname $DIR); pwd)
+test -z "$PYTHON_EXEC" && PYTHON_EXEC=${BASE_DIR}/python-env/bin/python
+test -z "$SUPER_USER" && SUPER_USER=postgres
+test -z "$MOX_DB" && MOX_DB=mox
+test -z "$MOX_DB_USER" && MOX_DB_USER=mox
+test -z "$MOX_DB_PASSWORD" && MOX_DB_PASSWORD=mox
+
+cd $DIR
+
 PYTHON=${PYTHON_EXEC}
 
 export PGPASSWORD="$MOX_DB_PASSWORD"
@@ -17,10 +27,13 @@ sudo -u postgres psql -c "GRANT ALL ON DATABASE $MOX_DB TO $MOX_DB_USER"
 sudo -u postgres psql -d $MOX_DB -f basis/dbserver_prep.sql
 
 # Setup AMQP server settings
-sudo -u $SUPER_USER psql -d $MOX_DB -c "insert into amqp.broker
+if test -n "$MOX_AMQP_HOST"
+then
+    sudo -u $SUPER_USER psql -d $MOX_DB -c "insert into amqp.broker
 (host, port, vhost, username, password)
 values ('$MOX_AMQP_HOST', $MOX_AMQP_PORT, '$MOX_AMQP_VHOST', '$MOX_AMQP_USER',
 '$MOX_AMQP_PASS');"
+fi
 
 # Grant mox user privileges to publish to AMQP
 sudo -u $SUPER_USER psql -d $MOX_DB -c "GRANT ALL PRIVILEGES ON SCHEMA amqp TO $MOX_DB_USER;
--- a/oio_rest/oio_rest/db.py
+++ b/oio_rest/oio_rest/db.py
@@ -11,11 +11,12 @@ from psycopg2.pool import ThreadedConnectionPool
 from jinja2 import Environment, FileSystemLoader
 from dateutil import parser as date_parser
 
-from .db_helpers import get_attribute_fields, get_attribute_names
-from .db_helpers import get_field_type, get_state_names, get_relation_field_type
-from .db_helpers import (Soegeord, OffentlighedUndtaget, JournalNotat,
-                        JournalDokument, DokumentVariantType, AktoerAttr,
-                        VaerdiRelationAttr)
+from .db_helpers import (
+    get_attribute_fields, get_attribute_names, get_field_type,
+    get_state_names, get_relation_field_type, Soegeord, OffentlighedUndtaget,
+    JournalNotat, JournalDokument, DokumentVariantType, AktoerAttr,
+    VaerdiRelationAttr,
+)
 
 from .authentication import get_authenticated_user
 
--- a/oio_rest/oio_rest/db.py
+++ b/oio_rest/oio_rest/db.py
@@ -1,6 +1,6 @@
+import datetime
+import enum
 import os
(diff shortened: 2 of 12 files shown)


danchr:
> Hvad med at oprette en sag på det i Redmine og gøre den til undersag af punkt-release-sagen?

[Sag 22658](https://redmine.magenta-aps.dk/issues/22658) oprettet, og PR justeret!

oio_rest/oio_rest/db.py
```diff
@@ -3,6 +3,8 @@
 from datetime import datetime, timedelta
 
 import psycopg2
+import flask
```
agger-magenta: Hvad med at gøre det mere eksplicit, hvad det drejer sig om, ved i stedet at skrive:

`from flask import g as global_context`

?

agger-magenta: Så er der ingen, der er i tvivl om, hvad der foregår nede i `get_connection`.

danchr: Dels fordi modul-imports er nemmere at resonere omkring, og fordi `flask.g` eller blot `g` er det egentlige navn.

oio_rest/oio_rest/db.py
```diff
@@ -35,33 +39,58 @@
...
-        password=settings.DB_PASSWORD,
-        host=getattr(settings, 'DB_HOST', 'localhost'),
-        port=getattr(settings, 'DB_PORT', 5432),
-    )
-    connection.autocommit = True
-    return connection
+    """Handle all intricacies of connecting to Postgres.
+
+    We stash the current connection using Flask's `application
+    globals`_, ``g``, to avoid continually re-estrablishing the
```
agger-magenta: "establishing"

oio_rest/requirements.txt
```diff
@@ -5,7 +5,7 @@ Jinja2==2.10
 MarkupSafe==1.0
 Werkzeug==0.14.1
 itsdangerous==0.24
-psycopg2-binary==2.7.4
+psycopg2cffi-compat
```
agger-magenta: Version?

danchr: Requiring an explicit version wouldn't really give us anything, since `psycopg2cffi-compat` itself [doesn't specify versions of either dependancy](https://github.com/dstufft/psycopg2cffi-compat/blob/master/setup.py#L22). Besides, the psycopg2 API is quite stable…

(Ideally, we should move to using something else, either `pipenv` or an additional constraints file, so that we can both record “known working” and the intention for the requirements.)

oio_rest/oio_rest/db.py
danchr: Thanks, fixed!

danchr:
@istherepie Could you perhaps take a look at this PR — particularly the changes to `initdb.sh`?

oio_rest/requirements.txt
agger-magenta: OK.

db/initdb.sh
```diff
@@ -4,6 +4,16 @@ set -b
...
 DIR=${DB_DIR}
 
+test -z "$DIR" && DIR=$(cd $(dirname $0); pwd)
+test -z "$BASE_DIR" && BASE_DIR=$(cd $(dirname $DIR); pwd)
+test -z "$PYTHON_EXEC" && PYTHON_EXEC=${BASE_DIR}/python-env/bin/python
+test -z "$SUPER_USER" && SUPER_USER=postgres
+test -z "$MOX_DB" && MOX_DB=mox
+test -z "$MOX_DB_USER" && MOX_DB_USER=mox
+test -z "$MOX_DB_PASSWORD" && MOX_DB_PASSWORD=mox
+
```
istherepie: Preferably the script should fail to execute rather than having fallback values, in a production setting the fallback values will not work - however its a lot better than it was before

oio_rest/oio_rest/db.py
```diff
@@ -35,33 +39,58 @@
...
-"""
+pool = ThreadedConnectionPool(
+    settings.DB_MIN_CONNECTIONS,
+    settings.DB_MAX_CONNECTIONS,
+    database=settings.DATABASE,
+    user=settings.DB_USER,
+    password=settings.DB_PASSWORD,
+    host=getattr(settings, 'DB_HOST', 'localhost'),
+    port=getattr(settings, 'DB_PORT', 5432),
+)
```
istherepie: I do not fully understand the need for this change to introduce connection pooling (and the added complexity) with the current state of the database layer and would recommend a single persistent connection for general queries. 
In my opinion pooling should not be a default but rather a choice for a subset of queries, e.g. a specific workload such as caching or bulk writes etc.

In the context of performance tuning there are other options to increase the performance such as optimizing the structure of the database model/relations, and alterting field types for better targeting.

oio_rest/settings.py
```diff
@@ -11,6 +11,12 @@
 DB_USER = getenv('DB_USER', 'mox')
 DB_PASSWORD = getenv('DB_PASS', 'mox')
 
+# Per-process limits on the amount of database connections. Setting
+# the minimum to a non-zero value ensures that the webapp opens this
+# amount at load, failing if the database isn't available.
+DB_MIN_CONNECTIONS = int(getenv('DB_MIN_CONNECTIONS', '0'))
+DB_MAX_CONNECTIONS = int(getenv('DB_MAX_CONNECTIONS', '10'))
```
istherepie: I've added recommendations against connection pooling, 
however I think that 10 connections are too many as a default either way for a general purpose workload.

PR #53, opened 2018-05-31
#22658 installer tweaks

danchr:
* Add an `nginx` instance to the default install, bind `gunicorn` to a socket rather than a port, and add rudimentary support for configuring SSL in `nginx`.
* Suppress a prompt from `apt-get` during the installation.
* Fix installer logging, and sort the keys in the grain output.
* Set the minimum amount of database connections for a slight speedup of imports, etc.
* Run gunicorn under the `mox` user rather than whomever happens to run the installer.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,6 +2,8 @@
 # DOES NOT AUTOMATICALLY INSTALL THE FOLLOWING COMPONENTS:
 # Mox Rest Frontend, Mox Advis, Mox Elk log
 
+set -e
+
 ## Check if executed as root
 if [ $( whoami ) == "root" ]; then
     echo "Do not run as root. We'll sudo when necessary"
--- a/install.sh
+++ b/install.sh
@@ -13,7 +13,7 @@ fi
 ## Variables
 
 # Base directory (current)
-BASE_DIR=$( pwd )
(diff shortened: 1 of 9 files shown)


agger-magenta:
Note that the web service should be served (i.e., gunicorn should be run) by a user that does not have write access to the Python code.

agger-magenta:
... I propose a user called, e.g., `mox_daemon` or `oio_rest` which has absolutely minimal permissions to run gunicorn.

istherepie:
Please note that the installer is NOT used in the automatic deployment procedure, it was added solely as a local install for development purposes, meaning that upcoming deployments will not benefit from any additions to the installer. 
As such it may be unnecessary to create service users and to install a proxy service/SSL support.

agger-magenta:
The point of having the nginx configuration is to have a developer friendly installation that is as close to the production setup as possible. This makes it easier to work with and to reason about how the system behaves.

danchr:
@agger-magenta Jeg har omplantet ændringer til spidsen af udviklingsgrenen, med lidt justeringer til c4e88ceff76cb670f170fa54f844bc5f1681d844 så vi rent faktisk viser alting. Kan du tage et nyt kig?

PR #55, opened 2018-06-01
setup.py & docs: obtain correct version from VERSION

danchr:
The previous logic was flawed, and used the wrong location, given that
we have a `VERSION` file for this specific case. While at it, I fixed
the authors line in `setup.py` to just state “Magenta ApS”, and added
a test to verify agreement between `NEWS.rst` and `VERSION`.

The motiviation for doing this as a direct hotfix/post-release is that
our current documentation gets the version wrong.


oio_rest/setup.py
```diff
@@ -2,13 +2,17 @@
...
 from setuptools import find_packages, setup
 
-version = '0.0.1'
-authors = 'C. Agger, Jørgen Ulrik B. Krag, Thomas Kristensen, Seth Yastrov'
 basedir = os.path.dirname(__file__)
 
+with open(os.path.join(basedir, '..', 'NEWS.rst')) as fp:
+    version = re.findall(r'^Version ([^,]*),', fp.read(), re.MULTILINE)[0]
+
+authors = 'C. Agger, Jørgen Ulrik B. Krag, Thomas Kristensen, Seth Yastrov'
```
agger-magenta: Dan, Christoffer og Steffen bør også inkluderes i "authors".

agger-magenta: Nu vi er ved det, altså.

doc/conf.py
```diff
@@ -60,11 +62,10 @@
 #
 # The full version, including alpha/beta/rc tags.
 with open('../NEWS.rst') as fp:
-  release = [l.split()[2].rstrip(',')
-             for l in fp if l.startswith('Version')][0]
+  release = re.findall(r'^Version ([^,]*),', fp.read(), re.MULTILINE)[0]
```
agger-magenta: Det er min opfattelse, at det er unødigt kompliceret og gør det let at introducere fejl at parse NEWS-filen for info om versions-nummer.

Det er det, VERSION-filen er der for. Som i øvrigt også burde opdateres.

Er denne ændring vigtig nok til at lave en minirelease nu uden at overveje de andre ting, der evt. hurtigt kunne fixes?

PR #61, opened 2018-08-23
Feature/19961 remove amqp from db layer

robertjensen:
This PR removes AMQP from the mox core and replaces the notification functionality with internal PostgreSQL notifications. This has a very positive influence on performance, simplifies the mox code and removes the dependency on a very specific fork of pg_amqp.

If AMQP messaging is wanted, this can be achieved by an external notification→AMQP relay. An example of such a relay is provided. 

The Postgres notifications could also be relayed to other messaging protocols or used directly in an external application, these examples are not provided.


python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+import psycopg2
+from os import getenv
+
+
+class Heartbeat(threading.Thread):
+    """ Heartbeat thread. Will ensure a heartbeat message approximately
+    every two seconds """
+    def __init__(self):
+        threading.Thread.__init__(self)
+        pika_params = pika.ConnectionParameters('localhost')
```
istherepie: AMQP host parameter is hard coded.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+        pika_connection = pika.BlockingConnection(pika_params)
+        self.amqp = pika_connection.channel()
+        self.amqp.queue_declare(queue='mox.heartbeat')
```
istherepie: AMQP queue name is hard coded.

robertjensen: This could in principle go into the configuration, but historically the documentation has always mentioned the hard-coded name 'mox.notifications', so I took it as a natural generalization to also hard-code the heartbeat name.

agger-magenta: I think we shouldn't hard code host and queue name. Before, when this was in the database, it wasn't easy to put in the onfiguration, I think, or at least we didn't do it. 

But it would be nice to be able to have a separate server & also to change the queue name.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+import threading
+import select
+import time
+import json
+import pika
+import psycopg2
+from os import getenv
+
+
+class Heartbeat(threading.Thread):
```
danchr: Like other thread class below, this class has two methods, one if which is `__init__`. That's a very good sign that it shouldn't, in fact be a class. Just use a function, and pass it to `threading.Thread`, if needs be.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+    """ Heartbeat thread. Will ensure a heartbeat message approximately
+    every two seconds """
+    def __init__(self):
+        threading.Thread.__init__(self)
```
danchr: Use `super().__init__()` instead.

python_agents/notification_service/notify_to_amqp_service.py
danchr: > AMQP host parameter is hard coded.

Agreed.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def run(self):
+        while True:
+            time.sleep(2)
+            self.amqp.basic_publish(exchange='',
+                                    routing_key='mox.heartbeat',
+                                    body=json.dumps(time.time()))
+
+
+class PgnotifyToAmqp(threading.Thread):
```
danchr: I'm not fond of this class name, as it seems a confused mix of uppercase and lowercase, and uses lowercase for some abbreviations usually written as uppercase. `AMQPNotifier` seems more apt and readable.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+                    objecktID = payload_dict['data'][objekttype + '_id']
+                    amqp_payload['objecktID'] = objecktID
+                    amqp_payload['objekttype'] = objekttype
+                    lc = payload_dict['data']['registrering']['livscykluskode']
+                    amqp_payload['livscykluskode'] = lc
+                    self.amqp.basic_publish(exchange='',
+                                            routing_key='mox.notifications',
+                                            body=json.dumps(amqp_payload))
+
+if __name__ == '__main__':
```
danchr: Use `click` to add support for command line arguments.

python_agents/notification_service/notify_to_amqp_service.py
danchr: Furthermore, the heartbeat idea is neat and useful for monitoring, but also more functionality that we need to test. It might be worthwhile to get a simple implementation in without it, and add it as a followup.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+                self.amqp.basic_publish(exchange='',
+                                        routing_key='mox.heartbeat',
+                                        body=json.dumps(time.time()))
+            else:
+                self.pg_conn.poll()
+                self.pg_conn.commit()
+                while self.pg_conn.notifies:
+                    notify = self.pg_conn.notifies.pop(0)
+                    payload_dict = json.loads(notify.payload)
+                    amqp_payload = {}
```
danchr: This section below is hard to read — please assign all simple values directly to the `dict` literal.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+                    table = payload_dict['table']
+                    objekttype = table[0:table.find('registrering')-1]
+                    amqp_payload['beskedtype'] = 'Notification'
+                    objecktID = payload_dict['data'][objekttype + '_id']
+                    amqp_payload['objecktID'] = objecktID
+                    amqp_payload['objekttype'] = objekttype
+                    lc = payload_dict['data']['registrering']['livscykluskode']
+                    amqp_payload['livscykluskode'] = lc
+                    self.amqp.basic_publish(exchange='',
```
danchr: This is a good place for a line break.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+                    notify = self.pg_conn.notifies.pop(0)
+                    payload_dict = json.loads(notify.payload)
+                    amqp_payload = {}
+                    table = payload_dict['table']
+                    objekttype = table[0:table.find('registrering')-1]
+                    amqp_payload['beskedtype'] = 'Notification'
+                    objecktID = payload_dict['data'][objekttype + '_id']
+                    amqp_payload['objecktID'] = objecktID
+                    amqp_payload['objekttype'] = objekttype
+                    lc = payload_dict['data']['registrering']['livscykluskode']
```
danchr: `lc` doesn't really say _life cycle code_ to me. If you want to keep the line short, consider either a `\` or wrapping the assignment in parentheses.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+        self.amqp.queue_declare(queue='mox.notifications')
+        self.amqp.queue_declare(queue='mox.heartbeat')
+
+    def run(self):
+        while True:
+            self.pg_cursor.execute("LISTEN mox_notifications;")
+            self.pg_conn.poll()
+            self.pg_conn.commit()
+            if select.select([self.pg_conn], [], [], 60) == ([], [], []):
+                print('timeout')
```
danchr: This is leftover debugging code — we should either get rid of it, or use a logger with a level of `DEBUG`.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+        self.pg_cursor = self.pg_conn.cursor()
+        pika_params = pika.ConnectionParameters('localhost')
+        pika_connection = pika.BlockingConnection(pika_params)
+        self.amqp = pika_connection.channel()
+        self.amqp.queue_declare(queue='mox.notifications')
+        self.amqp.queue_declare(queue='mox.heartbeat')
+
+    def run(self):
+        while True:
+            self.pg_cursor.execute("LISTEN mox_notifications;")
```
danchr: Do you really need to execute the listen call on every loop?

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+            self.pg_conn.poll()
+            self.pg_conn.commit()
+            if select.select([self.pg_conn], [], [], 60) == ([], [], []):
+                print('timeout')
+                self.amqp.basic_publish(exchange='',
```
danchr: You're sending a heatbeat from here — so why the other class?

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+                                        routing_key='mox.heartbeat',
+                                        body=json.dumps(time.time()))
+            else:
+                self.pg_conn.poll()
+                self.pg_conn.commit()
+                while self.pg_conn.notifies:
```
danchr: Please consider inserting a few extra line breaks here and there to separate sections of logic and thereby increasing readability — prior to the `while` is a good location, for example.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+    def run(self):
+        while True:
+            self.pg_cursor.execute("LISTEN mox_notifications;")
+            self.pg_conn.poll()
+            self.pg_conn.commit()
+            if select.select([self.pg_conn], [], [], 60) == ([], [], []):
+                print('timeout')
+                self.amqp.basic_publish(exchange='',
+                                        routing_key='mox.heartbeat',
+                                        body=json.dumps(time.time()))
```
danchr: Do you really need `json.dumps()` for converting a simple integer to a string? Why not `str()` — and is that even needed?

FWIW an ISO 8601 timestamp might be a better alternative, since that's readable by humans as well as machines.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -0,0 +1,95 @@
...
+        password=getenv("DB_PASS", "mox"),
+        host=getenv("DB_HOST", "localhost")
+    )
+    heartbeat = Heartbeat()
+    heartbeat.daemon = True
+    notify2amqp.start()
+    heartbeat.start()
+
+    while notify2amqp.is_alive():
+        time.sleep(2)
```
danchr: You've started two threads, only to actively wait for one of them to succeed, while the other one is a daemon to ensure that you're _not_ waiting for it. That's an awful lot of complexity, for something that doesn't _really_ need threads.

Furthermore, your heartbeat thread doesn't _necessarily_ do what you want it to do — should the other thread stall rather than terminate, it'll keep sending out heartbeats. And given that you're also doing heartbeats from the notification thread, you don't need that one.

Just do everything on the main thread instead.

python_agents/notification_service/notify_to_amqp_service.py
agger-magenta: And also, it's a class which has only two members, one of which is `__init__`, and the other of which is called `run()` - as Dan has pointed out, this is not really a class but a function. cf. https://www.youtube.com/watch?v=o9pEzgHorH0 which is worthwhile if you have the time.

python_agents/notification_service/notify_to_amqp_service.py
agger-magenta: I agree; since we don't really need threads, we should avoid them.

python_agents/notification_service/notify_to_amqp_service.py
agger-magenta: I agree! Only please don't use a "\". Backspace continuations are brittle and, as a general rule, shouldn't be used.

python_agents/notification_service/notify_to_amqp_service.py
danchr: Indeed, if you want a bit of logic simplification, you could create two functions: a generator to extract notifications, and a consumer for sending them to AMQP.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -9,25 +9,21 @@
...
-        self.amqp = pika_connection.channel()
-        self.amqp.queue_declare(queue='mox.heartbeat')
-
-    def run(self):
-        while True:
-            time.sleep(2)
-            self.amqp.basic_publish(exchange='',
-                                    routing_key='mox.heartbeat',
-                                    body=json.dumps(time.time()))
+def Heartbeat(amqp_notifier):
```
danchr: Function names should not start with an uppercase letter.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -9,25 +9,21 @@
...
+    pika_params = pika.ConnectionParameters('localhost')
+    pika_connection = pika.BlockingConnection(pika_params)
+    amqp = pika_connection.channel()
+    amqp.queue_declare(queue='mox.heartbeat')
+    while amqp_notifier.ttl > 0:  # Ensure AMQPNotifier is running
+        amqp_notifier.ttl -= 1
+        print(amqp_notifier.ttl)
```
danchr: Leftover debugging code.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -9,25 +9,21 @@
...
-            time.sleep(2)
-            self.amqp.basic_publish(exchange='',
-                                    routing_key='mox.heartbeat',
-                                    body=json.dumps(time.time()))
+def Heartbeat(amqp_notifier):
+    pika_params = pika.ConnectionParameters('localhost')
+    pika_connection = pika.BlockingConnection(pika_params)
+    amqp = pika_connection.channel()
+    amqp.queue_declare(queue='mox.heartbeat')
+    while amqp_notifier.ttl > 0:  # Ensure AMQPNotifier is running
```
danchr: Insert newline.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -9,25 +9,21 @@
...
+        amqp_notifier.ttl -= 1
+        print(amqp_notifier.ttl)
+        time.sleep(2)
+        amqp.basic_publish(exchange='',
```
danchr: Insert newline.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -47,49 +43,53 @@ def __init__(self, database, user, password, host):
...
                     objecktID = payload_dict['data'][objekttype + '_id']
-                    amqp_payload['objecktID'] = objecktID
-                    amqp_payload['objekttype'] = objekttype
-                    lc = payload_dict['data']['registrering']['livscykluskode']
-                    amqp_payload['livscykluskode'] = lc
+                    registrering = payload_dict['data']['registrering']
+                    livscykluskode = registrering['livscykluskode']
+
+                    amqp_payload = {'beskedtype': 'Notification',
+                                    'objecktID': objecktID,
```
danchr: Typo: “objecktID” is neither English nor Danish.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -47,49 +43,53 @@ def __init__(self, database, user, password, host):
...
                     table = payload_dict['table']
                     objekttype = table[0:table.find('registrering')-1]
-                    amqp_payload['beskedtype'] = 'Notification'
                     objecktID = payload_dict['data'][objekttype + '_id']
-                    amqp_payload['objecktID'] = objecktID
-                    amqp_payload['objekttype'] = objekttype
-                    lc = payload_dict['data']['registrering']['livscykluskode']
-                    amqp_payload['livscykluskode'] = lc
+                    registrering = payload_dict['data']['registrering']
+                    livscykluskode = registrering['livscykluskode']
```
danchr: Is there any particular need for this variable assignment?

python_agents/notification_service/notify_to_amqp_service.py
danchr: The above comment still stands; please use `click`.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -47,49 +43,53 @@ def __init__(self, database, user, password, host):
...
-    notify2amqp = PgnotifyToAmqp(
+    amqp_notifier = AMQPNotifier(
         database=getenv("DB_NAME", "mox"),
         user=getenv("DB_USER", "mox"),
         password=getenv("DB_PASS", "mox"),
         host=getenv("DB_HOST", "localhost")
     )
-    heartbeat = Heartbeat()
+
+    heartbeat = threading.Thread(target=Heartbeat, args=[amqp_notifier])
```
danchr: No — this is a complete and utter waste of resources, and adds complexity for no benefit whatsoever.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -9,25 +9,21 @@
...
+        amqp_notifier.ttl -= 1
+        print(amqp_notifier.ttl)
+        time.sleep(2)
+        amqp.basic_publish(exchange='',
+                           routing_key='mox.heartbeat',
+                           body=str(time.time()))
 
 
-class PgnotifyToAmqp(threading.Thread):
+class AMQPNotifier(threading.Thread):
```
danchr: Get rid of this class.

python_agents/notification_service/notify_to_amqp_service.py
```diff
@@ -47,49 +43,53 @@ def __init__(self, database, user, password, host):
...
             self.pg_conn.poll()
             self.pg_conn.commit()
             if select.select([self.pg_conn], [], [], 60) == ([], [], []):
-                print('timeout')
                 self.amqp.basic_publish(exchange='',
                                         routing_key='mox.heartbeat',
-                                        body=json.dumps(time.time()))
+                                        body=str(time.time()))
             else:
                 self.pg_conn.poll()
```
danchr: What does this poll do? Shouldn't you do something with the result?

PR #63, opened 2018-08-27
Added support for handling SAML2 assertions already wrapped in respon…

cmoesgaard:
…ses (as is the case with the SSO response)

### DIFF
--- a/oio_rest/oio_rest/auth/saml2.py
+++ b/oio_rest/oio_rest/auth/saml2.py
@@ -34,15 +34,22 @@ class Saml2_Assertion(OneLogin_Saml2_Response):
         self.idp_entity_id = idp_entity_id
         self.idp_cert = idp_cert
 
-        # OneLogin's methods expect the data wrapped in a Response element,
-        # so we fake it here.
-        self.response = """
-        <samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
-                        xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">
-            %s
-        </samlp:Response>""" % assertion_xml
+        document = fromstring(assertion_xml)
+        if 'Response' not in document.tag:
+            # OneLogin's methods expect the data wrapped in a Response element,
+            # so we fake it here.
+            response = """
+            <samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
+                            xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">
+                %s
+            </samlp:Response>""" % assertion_xml
 
-        self.original_document = fromstring(self.response)
+            document = fromstring(response)
+            self.response = response.encode('ascii')
+        else:
+            self.response = assertion_xml
+
+        self.original_document = document
 
         super(Saml2_Assertion, self).__init__(
             OneLogin_Saml2_Settings(
@@ -60,7 +67,7 @@ class Saml2_Assertion(OneLogin_Saml2_Response):
                 },
                 sp_validation_only=True,
             ),
-            base64.b64encode(self.response.encode('ascii')),
+            base64.b64encode(self.response),
         )
 
     def __query(self, query):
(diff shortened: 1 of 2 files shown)


oio_rest/oio_rest/auth/saml2.py
```diff
@@ -34,15 +34,22 @@ def __init__(self, assertion_xml, mox_entity_id, idp_entity_id, idp_url,
...
-        # so we fake it here.
-        self.response = """
-        <samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
-                        xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">
-            %s
-        </samlp:Response>""" % assertion_xml
-
-        self.original_document = fromstring(self.response)
+        document = fromstring(assertion_xml)
+        if 'Response' not in document.tag:
```
danchr: Any particular reason not do a direct comparison?

oio_rest/oio_rest/auth/saml2.py
```diff
@@ -34,15 +34,22 @@ def __init__(self, assertion_xml, mox_entity_id, idp_entity_id, idp_url,
...
+            # OneLogin's methods expect the data wrapped in a Response element,
+            # so we fake it here.
+            response = """
+            <samlp:Response xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
+                            xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion">
+                %s
+            </samlp:Response>""" % assertion_xml
```
agger-magenta: I think we should use `format()` for string formatting, it's more powerful and easier to read.

danchr: In @cmoesgaard's defence, he merely copied this code from elsewhere 🙂 In this case, though, I'd say that `''.join(…)` is customary, given that the code isn't doing any actual formatting.

agger-magenta: Whatever! I'd like to deprecate the `%` operator for string formatting.

agger-magenta: And in Python 3.6 we'll get these cute `f"..."` thingies.

oio_rest/oio_rest/auth/saml2.py
cmoesgaard: `document.tag` contains the tag's namespace as well, e.g. `'{urn:oasis:names:tc:SAML:2.0:protocol}Response'`

As an alternative, we could so some simple regex matching to extract the part after the `{}` part, and do our comparison on that.

oio_rest/oio_rest/auth/saml2.py
cmoesgaard: Agreed, I'll fix that.

oio_rest/oio_rest/auth/saml2.py
danchr: Gah, and there's no way to access the unprefixed tag, or add a namespace? (I remember running into this earlier, and yeah, it _is_ bloody annoying…)

cmoesgaard: No way of getting _just_ the tag name, as far as I can tell.

cmoesgaard: It probably isn't going to change, no. You're right.

agger-magenta: I think either that or add a comment on why we're doing it like this.

oio_rest/oio_rest/auth/saml2.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, assertion_xml, mox_entity_id, idp_entity_id, idp_url,
         self.idp_cert = idp_cert
 
         document = fromstring(assertion_xml)
-        if 'Response' not in document.tag:
+        if document.tag != '{urn:oasis:names:tc:SAML:2.0:protocol}Response':
```
danchr: This still feels vaguely horrible… that's definitely not your fault, though, but rather a case of weird APIs for a horrible format.

PR #68, opened 2018-09-13
[RFC] add basic support for counting objects

danchr:
This is a quick-and-dirty implementation which is significantly faster in
most cases. The main downside is that it doesn't support restrictions,
yet, and also doesn't consider effect and registration time beyond now.

Posted here for comments and feedback.

Some benchmarks in a database with ~56k objects, all valid with
searches that yield ~800 objects:

- filtering for validity, an attribute and one relation:
  from ~2.7s to <0.1s
- filtering for validity, an attribute and two relations:
  from ~4.8s to ~4.5s

The extra relation happens to be largely redundant in the MO data
model, but the cause of the significant overhead is likely joining the
relations table twice. If we could get rid of that, near-instant
searches seem doable.

The motivation for doing this is that the MO front page took more than 30s
to render with a large™ data set.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -66,7 +66,7 @@ pipeline {
   post {
     always {
       junit healthScaleFactor: 200.0,           \
-        testResults: '**/build/reports/*.xml'
+        testResults: 'oio_rest/TEST-*.xml'
 
       warnings canRunOnFailed: true, consoleParsers: [
         [parserName: 'Sphinx-build'],
@@ -75,6 +75,9 @@ pipeline {
 
       cobertura coberturaReportFile: 'oio_rest/coverage.xml',    \
         maxNumberOfBuilds: 0
+
+      junit healthScaleFactor: 200.0,     \
+        testResults: 'mora/build/reports/*.xml'
 
       cleanWs()
     }
--- a/oio_rest/oio_rest/db.py
+++ b/oio_rest/oio_rest/db.py
@@ -810,6 +810,38 @@ def search_objects(class_name, uuid, registration,
     return output
 
 
+def count_objects(class_name, registration):
+    for groupname, group in registration.items():
+        if isinstance(group, dict):
+            for values in group.values():
+                for value in values:
+                    assert value['virkning'] is None, value.get('virkning')
+
+                    del value['virkning']
+        else:
+            # TODO :(
+            raise NotImplementedError('cannot handle ' + groupname)
+
+    sql_template = jinja_env.get_template('count_objects.sql')
+
+    sql = sql_template.render(
+        cls=class_name,
+        reg=registration,
+    )
+
+    with get_connection() as conn, conn.cursor() as cursor:
+        try:
+            cursor.execute(sql)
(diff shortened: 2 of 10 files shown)


oio_rest/oio_rest/db.py
```diff
@@ -810,6 +810,38 @@ def search_objects(class_name, uuid, registration,
     return output
 
 
+def count_objects(class_name, registration):
```
agger-magenta: The registration that you specify is obviusly meant as a scope for the counting, i.e. conditions to be applied to the select statement.

In that case, to lessen confusion, it might be called that, "scope" e.g., and a docstring could specify that this is supposed to be a registration, i.e. a dict with attributes, states and relations?

oio_rest/requirements-test.txt
```diff
@@ -1,8 +1,7 @@
 -r requirements.txt
 Flask-Testing
 testing.postgresql
-pytest
-pytest-cov
+unittest-xml-reporting
```
agger-magenta: Once again, is this change related to this pull request?

oio_rest/oio_rest/oio_rest.py
```diff
@@ -236,6 +236,14 @@ def get_objects(cls):
 
         valid_list_args = TEMPORALITY_PARAMS | {'uuid'}
 
+        if args.get('count') and args['count'] == '1':
```
agger-magenta: :+1: 

The API documentation should be updated, though.

oio_rest/oio_rest/db.py
danchr: > The registration that you specify is obviusly meant as a scope for the counting, i.e. conditions to be applied to the select statement.
> 
> In that case, to lessen confusion, it might be called that, "scope" e.g., and a docstring could specify that this is supposed to be a registration, i.e. a dict with attributes, states and relations?

I'd be inclined to agree, but that's what the other functions call it 🤷‍♂️

oio_rest/oio_rest/db.py
```diff
@@ -810,6 +810,38 @@ def search_objects(class_name, uuid, registration,
...
+    for groupname, group in registration.items():
+        if isinstance(group, dict):
+            for values in group.values():
+                for value in values:
+                    assert value['virkning'] is None, value.get('virkning')
```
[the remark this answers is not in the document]
danchr: A sanity check; if it's something else, someone is doing something horribly wrong. It might be related to the test failures in `dokument`, though…

Jenkinsfile
```diff
@@ -76,6 +76,9 @@ pipeline {
       cobertura coberturaReportFile: 'oio_rest/coverage.xml',    \
         maxNumberOfBuilds: 0
 
+      junit healthScaleFactor: 200.0,     \
```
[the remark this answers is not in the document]
danchr: Yeah, it should be a separate PR, but I wanted to be able to get subtests on Jenkins — and this change was mostly copied from MO so I could get that. It's an RFC, not a final PR ;)

PR #69, opened 2018-09-19
Remove patch system

carlbordum:
First attempt to remove the patch system

Its quite a big diff, so I recommend you avoid Githubs diff tool.

Instead, you can use this command which compares the generated files with those 1 commit older than my work: `git diff --minimal --diff-filter=d --ignore-blank-lines -b -w cc047ad8 6e559cf2 db/db-templating/generated-files/*`

Most of those diffs show changes in comments, so I've written a small tool that removes comments and redundant whitespace and prints the diffs then. Place it in `db/db-templating/` before executing it with Python 3.

PL/pgSQL note:
Inline comments start with 2 hyphens.
Multiline comments are between /* and */.

I have assumed that all files in `db/db-templating/patches` has to go. Furthermore, I removed:
`db/db-templating/apply-template.py`
`db/db-templating/generate-sql-tbls-types-funcs-for-oiotype.sh`
`db/db-templating/generate_diff_files_helper.sh`

The Jinja templates can still get a lot prettier. Focus in this pull request is to get rid of the patch system with the smallest reasonable diff.

Script for diffing without pl/pgsql comments:
``` python
import os
from pathlib import Path
import re
import subprocess

COMMIT_BEFORE = "63a41e638f4f2b3792da44eb050608a939667f6d"
COMMIT_AFTER = "3b5386a5056d6f6ebde676d11d671ace6a1c1b10"


def remove_multiline_comments(text):
    return re.sub(r"/\*.*?\*/", "", text, flags=re.DOTALL)


def remove_inline_comments(text):
    return re.sub(r"--.*", "", text)


def remove_comments(text):
    return remove_multiline_comments(remove_inline_comments(text))


def remove_redundant_spacing(text):
    # only between words; we let `diff` deal with indentation.
    # if we remove newlines, the produced diff is ugly (and much slower too).
    return re.sub(r"\n+", "\n", re.sub(r"(?<=\w)\s+(?=\w)", " ", text))


def trim_document(document):
    return remove_redundant_spacing(remove_comments(document))


if __name__ == "__main__":
    precommit = (
        subprocess.check_output(["git", "rev-parse", "HEAD"]).decode("utf-8").strip()
    )
    gendir = Path("generated-files/")

    for commit in (COMMIT_BEFORE, COMMIT_AFTER):
        subprocess.run(["git", "checkout", commit])

        stripped_files = []
        for filename in os.listdir(gendir):
            with open(gendir / filename) as f:
                content = f.read()
            stripped_files.append((filename, trim_document(content)))

        outdir = Path("generated-files-%s" % commit)
        outdir.mkdir(exist_ok=True)

        for filename, content in stripped_files:
            with open(outdir / filename, "w+") as f:
                f.write(content)

    subprocess.run(["git", "checkout", precommit])

    for filename in os.listdir(outdir):
        command = [
            "diff",
            "--color",
            "--ignore-space-change",
            "--ignore-trailing-space",
            "--ignore-all-space",
            "--ignore-blank-lines",
            "--strip-trailing-cr",
            "generated-files-%s/%s" % (COMMIT_BEFORE, filename),
            "generated-files-%s/%s" % (COMMIT_AFTER, filename),
        ]
        try:
            subprocess.check_call(command, stdout=subprocess.PIPE)
        except subprocess.CalledProcessError:
            print("\n", filename)
            subprocess.run(command)
```


danchr:
Seems like the tests fail on Jenkins — could you take a look at that?

carlbordum:
Forgot to mention: the diffs produced by the script are all changes in logical redundant parentheses, e.g. `((a.b) == c)` instead of `(a.b == c)`.

carlbordum:
Yes it is feasible

The diffs are only for 3 files, where 1, 2 and 3 lines have seen changes in parentheses, respectively.

This is because 3 patches changed the parentheses for just those files and not their siblings, so the fix (to avoid this diff) would be super ugly.

danchr:
> The diffs are only for 3 files, where 1, 2 and 3 lines have seen changes in parentheses, respectively.
> 
> This is because 3 patches changed the parentheses for just those files and not their siblings, so the fix (to avoid this diff) would be super ugly.

Gah, that does sound rather horrible — okay, let's leave it as-is until we start the proper review! (That is, when the tests pass 😈 )

agger-magenta:
I agree with Dan! We're not always so strict about tests because some tests might fail because of artefacts and expectations rather than actual errors, but in this case we really need to generate SQL which is identical or at least completely equivalent to what is done by the patch system, so when all the tests pass really is the correct moment to consider this task "reviewable".

db/db-templating/generated-files/as_search_sag.sql
```diff
@@ -669,72 +680,75 @@ ELSE
...
-									(
-										((relationTypeObj.journalDokumentAttr).offentlighedundtaget).Hjemmel IS NULL
-										OR
-										((a.journal_dokument_attr).offentlighedundtaget).Hjemmel ILIKE ((relationTypeObj.journalDokumentAttr).offentlighedundtaget).Hjemmel
-									)
-								)
-						)
-					)
-				)
+
```
agger-magenta: Why the change in formatting? Could it be avoided?

db/db-templating/generated-files/as_search_tilstand.sql
```diff
@@ -742,31 +753,34 @@ ELSE
...
-					(
-						(relationTypeObj.tilstandsVaerdiAttr).forventet IS NULL
-						OR
-						(a.tilstand_vaerdi_attr).forventet = (relationTypeObj.tilstandsVaerdiAttr).forventet
-					)
-		
-				)
-				)
+                
+
```
agger-magenta: As above (obviously, it's the same thing).

db/db-templating/templates/as_update.jinja.sql
```diff
@@ -78,38 +133,252 @@ IF relationer IS NOT NULL AND coalesce(array_length(relationer,1),0)=0 THEN
 ELSE
 
   --1) Insert relations given as part of this update
-  --2) Insert relations of previous registration, taking overlapping virknings into consideration (using function subtract_tstzrange)
+  --2) for aktivitet: Insert relations of previous registration, with index values not included in this update. Please notice that for the logic to work,
```
agger-magenta: This would be: For aktivitet, indsats, sag and tilstand, right?

All of those have indices for to-many relations.

agger-magenta:
A more general comment is that you shouldn't work on development, not even in your own repository, but on a feature branh with the ticket number included. But that's OK for this one, I think.

PR #70, opened 2018-10-05
Added notice about timezones in documentation on testing

carlbordum:
Hopefully the next dev won't waste time on this :)

### DIFF
--- a/doc/API-NOTES.rst
+++ b/doc/API-NOTES.rst
@@ -116,8 +116,8 @@ Make sure the parameter ``moxrestfrontend.rest.host`` in
 ``http://localhost:5000``.
 
 Make sure to set the local time to the same as Europe/Copenhagen when testing.
-This can be done with the following command: ``cp
-/usr/share/zoneinfo/Europe/Copenhagen /etc/localtime``.
+This can be done with the following command: ``sudo timedatectl set-timezone
+ Europe/Copenhagen``.
 
 Start the (AMQP) MOX REST frontend agent::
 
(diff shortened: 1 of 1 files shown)


doc/API-NOTES.rst
```diff
@@ -115,6 +115,10 @@ Make sure the parameter ``moxrestfrontend.rest.host`` in
 ``agents/MoxRestFrontend/moxrestfrontend.conf`` is set to
 ``http://localhost:5000``.
 
+Make sure to set the local time to the same as Copenhagen when testing. This
```
danchr: `Copenhagen` ➡️ `Europe/Copenhagen`.

doc/API-NOTES.rst
```diff
@@ -115,6 +115,10 @@ Make sure the parameter ``moxrestfrontend.rest.host`` in
...
+can be done with the following command: ``cp
+/usr/share/zoneinfo/Europe/Copenhagen /etc/localtime``.
```
danchr: Actually, `dpkg-reconfigure tzdata` is vaguely better IMO — but perhaps we should add a test to verify that this is indeed the case, or make the test check so in its `setUp` routine.

carlbordum: Advantage by copying the file is that its easier to put in a shell script.

Do you want me to change it?

agger-magenta: The correct/authorized way of doing this is:
```
sudo timedatectl set-timezone Europe/Copenhagen
```
That's better than copying files around yourself.

PR #72, opened 2018-10-10
Make `interface_test/test_dokument.sh` work.

carlbordum:
This test had some issues, because it was getting outdated and didnt
handle current mox responses correctly. This is the first step towards
redmine issue #24424 and updating it spawned 4 bug reports, opened as
sub-issues. Failing tests (reported bugs) have been out-commented with
mentions of their respective redmine issues.

### DIFF
--- a/interface_test/test_dokument.sh
+++ b/interface_test/test_dokument.sh
@@ -26,7 +26,7 @@ result=$(curl -s -S -k -X POST \
     -F "del_indhold3=@$DIR/test_data/test.xls" \
     $HOST_URL/dokument/dokument)
 
-uuid=$(echo $result | python3 -c "print(eval(input())['uuid'])")
+uuid=$(echo $result | python3 -c "import json, sys; print(json.load(sys.stdin)['uuid'])")
 
 
 if [ ! -z "$uuid" ]
--- a/interface_test/test_dokument.sh
+++ b/interface_test/test_dokument.sh
@@ -19,25 +19,28 @@
 DIR=$(dirname ${BASH_SOURCE[0]})
 source $DIR/config.sh
 
-result=$(curl -k -X POST \
+result=$(curl -s -S -k -X POST \
     -F "json=$(cat $DIR/test_data/dokument_opret.json)" \
     -F "del_indhold1=@$DIR/test_data/test.txt" \
     -F "del_indhold2=@$DIR/test_data/test.docx" \
     -F "del_indhold3=@$DIR/test_data/test.xls" \
     $HOST_URL/dokument/dokument)
-uuid=$(expr "$result" : '.*"uuid": "\([^"]*\)"')
 
-if [ ! -z $uuid ]
+uuid=$(echo $result | python3 -c "print(eval(input())['uuid'])")
+
+
+if [ ! -z "$uuid" ]
 then
     printf "\nOprettet dokument: $uuid"
 else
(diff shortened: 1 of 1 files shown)


interface_test/test_dokument.sh
```diff
@@ -19,25 +19,28 @@
...
+result=$(curl -s -S -k -X POST \
     -F "json=$(cat $DIR/test_data/dokument_opret.json)" \
     -F "del_indhold1=@$DIR/test_data/test.txt" \
     -F "del_indhold2=@$DIR/test_data/test.docx" \
     -F "del_indhold3=@$DIR/test_data/test.xls" \
     $HOST_URL/dokument/dokument)
-uuid=$(expr "$result" : '.*"uuid": "\([^"]*\)"')
 
-if [ ! -z $uuid ]
+uuid=$(echo $result | python3 -c "print(eval(input())['uuid'])")
```
danchr: `eval` is evil 😉 If the input is JSON, use:

```
python3 -c 'import json, sys; print(json.load(sys.stdin)["uuid"])'
```

Just to be safe…

interface_test/test_dokument.sh
```diff
@@ -19,25 +19,28 @@
...
     printf "\nOpret dokument fejlet: $result\n"
-    exit
+    exit 1
 fi
 
+
 # Import
 import_uuid=$(uuidgen)
-curl -k --write-out %{http_code} --output /tmp/dokument_opret.txt -X PUT \
+curl -s -S -k --output /tmp/dokument_opret.txt -X PUT \
```
danchr: Why not use `-f` if you want `curl` to fail on errors?

interface_test/test_dokument.sh
```diff
@@ -46,143 +49,170 @@ curl -k --write-out %{http_code} --output /tmp/dokument_opret.txt -X PUT \
...
-    printf "\nSearch del 1 successful"
-else
-    printf "\nError in search del 1.\n"
-    exit
-fi
+# DOESNT ACCEPT mimetype PARAMETER
+# redmine sag #24631
+
+printf "\nWarning: did not run 'search del 1' test."
+#if $(curl -k -s -S -H "Content-Type: application/json" "$HOST_URL/dokument/dokument?varianttekst=PDF&deltekst=doc_deltekst1A&mimetype=text/plain&uuid=$import_uuid" | grep -qi "$import_uuid")
```
agger-magenta: This test used to run fine on the old version of LoRa. Why not investigate why it is that `variant`, `mimetype` etc. don't work rather than comment out the code that uses them?

danchr: I'd say the likely cause is overly agressive validation of URL parameters. I've run into that elsewhere; for example, we no longer allow searches by object type.

agger-magenta: Exactly, we need to reallow that.

carlbordum: I did investigate. Let me know, if I'm wrong but it seems that these parameters actually never really worked, just that they were ignored because we didnt have the same input validation (and now they are getting rejected). With these changes, at least you can run the tests locally.

What do you think?

carlbordum: Sorry, my previous comment was posted without refreshing the browser.

Why is it good to allow them if we just ignore them?

carlbordum:
Resolved by #75.

PR #74, opened 2018-10-31
Updated to use flask_saml_sso for auth flow

cmoesgaard:
The build currently fails because we currently install `flask_saml_sso` directly from Github with the `-e` flag, which isn't supported in `setup.py`.
The solution to this will be to publish `flask_saml_sso` to PyPI, which will happen once https://github.com/magenta-aps/flask_saml_sso/pull/1 is merged.

### DIFF
--- a/oio_rest/oio_rest/app.py
+++ b/oio_rest/oio_rest/app.py
@@ -20,6 +20,8 @@ from .custom_exceptions import OIOFlaskException, AuthorizationFailedException
 from .custom_exceptions import BadRequestException
 from .auth import tokens
 
+import flask_saml_sso
+
 import settings
 
 app = Flask(__name__)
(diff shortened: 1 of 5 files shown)


oio_rest/oio_rest/app.py
```diff
@@ -57,6 +59,9 @@ def __init__(self, url_map, *items):
 indsats.IndsatsHierarki.setup_api(base_url=settings.BASE_URL, flask=app)
 tilstand.TilstandsHierarki.setup_api(base_url=settings.BASE_URL, flask=app)
 
+app.config.from_object(settings)
```
danchr: Longer term, we should make the same change to `oio_rest` as we did to `mora`, where we use a factory function & blueprints. I find it much nicer overall…

cmoesgaard: Yes, I agree.

PR #75, opened 2018-11-02
Feature/move-interface-tests

carlbordum:
Alright, the shell scripts and the test data in `interface_test` have been moved to our python test suite.

The only issues was with the dokument and sag endpoints, but they should be immensely easier to fix now. Just uncomment the tests one at a time :)


oio_rest/tests/test_rest_interface.py
```diff
@@ -0,0 +1,668 @@
+import pytest
```
danchr: On a high level, I'd prefer every single object type — such as `dokument` — to be a separate test file. In addition, you have some cases where you've commented out failing tests; I'd rather we use the `@unittest.expectedFailure` decorator, if possible.

danchr: Also; this particular import isn't used AFAICT.

oio_rest/tests/test_rest_dokument.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+class TestDokument(test_support.TestRestInterface):
+    @unittest.skip("I don't know what should happen here.")
+    def test_create_dokument_empty_dict(self):
+        result = self.client.post(
+            "/dokument/dokument",
+            data={"json": "{}"},
+        ).get_json()
+        assert result["uuid"] is not None
```
danchr: Please use the relevant `unittest.TestCase` methods:

```suggestion
        self.assertIsNotNone(result["uuid"])
```

oio_rest/tests/test_rest_dokument.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def test_create_dokument_missing_files(self):
+        result = self.client.post(
+            "/dokument/dokument",
+            data={
+                "json": open("tests/fixtures/dokument_opret.json", "rb").read(),
+            }
+        ).get_json()
+        assert "uuid" not in result and result["message"]
```
danchr: Likewise:

```suggestion
        self.assertNotIn("uuid", result)
        self.assertTrue(result["message"])
```

…or perhaps something even more thorough.

Jenkinsfile
```diff
@@ -4,7 +4,8 @@ pipeline {
   agent any
 
   environment {
-    PYTEST_ADDOPTS = '--color=yes'
+    PYTEST_ADDOPTS     = '--color=yes'
+    FILE_UPLOAD_FOLDER = './mox-upload'
```
danchr: You should set this in the test runner, specifically in either each `TestCase` or in the shared superclasses in `tests/util.py`.

The tests should not rely on the Jenkins infrastructure, and developers should not need to perform any modifications to their environment prior to running.

oio_rest/tests/test_validate.py
```diff
@@ -82,13 +82,12 @@ def setUp(self):
 
     def _json_to_dict(self, filename):
         """
-        Load a JSON file from /interface_test/test_data and return it as JSON
+        Load a JSON file from tests/fixtures and return it as JSON
```
danchr:
```suggestion
        Load a JSON file from ``tests/fixtures`` and return it as JSON
```

…also, not your fault, but directives like `:param:` require an empty line separating them from the preceding paragraph.

oio_rest/tests/test_rest_dokument.py
```diff
@@ -0,0 +1,239 @@
...
+
+    @unittest.expectedFailure
+    def test_dokument(self):
+        with self.subTest("Upload dokument"):
```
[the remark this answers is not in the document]
carlbordum: I get your point. However, in unittest multiple subtests at the same level rely on the previous to complete, but this way you at least know that this particular chunk of code is meant to upload a dokument.

Do you want me to change it?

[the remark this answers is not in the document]
danchr: In this particular instance, I think a separate `TestCase` with a `setUp` routine might make more sense. That'd also make it clearer which tests actually fail.

oio_rest/tests/test_rest_indsats.py
```diff
@@ -0,0 +1,29 @@
...
+            data={
+                "json": open("tests/fixtures/indsats_opret.json", "rt").read(),
+            },
+        )
+        self.assertEqual(result.status_code, 201)
+        uuid_ = result.get_json()["uuid"]
+        self.assertTrue(is_uuid(uuid_))
+
+    def test_indsats_put(self):
+        uuid_ = str(uuid.uuid4())
```
danchr: Just a note: Each of these tests should run with an ‘empty’ database. Hence, you can quite safely hard-code UUIDs within tests, and assume that your creation works as expected. My personal favourite is `00000000-0000-0000-0000-000000000000` :)

PR #81, opened 2018-11-08
Feature/24423_jenkins_db_tests

carlbordum:


oio_rest/oio_rest/utils/test_support.py
```diff
@@ -63,16 +160,7 @@ def _initdb():
...
-        [os.path.join(BASE_DIR, '..', 'db', 'mkdb.sh')],
-        env=env,
-        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
-        stdin=subprocess.DEVNULL,
-    ) as proc:
-        sql, errortext = proc.communicate()
-
-    assert proc.returncode == 0, 'mkdb failed:\n\n' + errortext.decode()
-
+    sql = _get_db_setup_sql(settings.DATABASE, settings.DB_USER)
```
danchr: 🎉 

This is precisely what we discussed at the meeting — that should make @jgj-magenta-aps's life easier!

oio_rest/oio_rest/utils/test_support.py
```diff
@@ -41,6 +41,103 @@ def psql():
     return psql
 
 
+def _get_db_setup_sql(db_name, db_user):
```
danchr: This is indeed quite a mess, as you note, and after reading the code, I do understand why. It's better than the _previous_ mess, though!

oio_rest/oio_rest/utils/test_support.py
```diff
@@ -41,6 +41,103 @@ def psql():
...
+    """Return the postgresql + pl/pgsql code necessary for our database
+    to work.
+
+    Brave souls can find db/mkdb.sh in the git history to see what this
```
danchr: Hey, I wrote that script — and you're right! 😁

oio_rest/oio_rest/utils/test_support.py
jgj-magenta-aps: jgj-magenta-aps is pleased with having such friendly peers

PR #87, opened 2018-11-14
pgtap is only required when testing

carlbordum:
Also don't load test functions in production database


oio_rest/oio_rest/utils/test_support.py
```diff
@@ -59,6 +59,7 @@ def _get_db_setup_sql(db_name, db_user):
     ALTER DATABASE "{db_name}" SET INTERVALSTYLE to 'sql_standard';
 
     CREATE SCHEMA test AUTHORIZATION "{db_user}";
+    CREATE EXTENSION IF NOT EXISTS "pgtap";
```
danchr: Erm, could we perhaps delay this to the test that runs the test functions? Otherwise, `full-run` would still break w/o `pgtap`.

PR #89, opened 2018-11-26
#25308 psql test cleanups

danchr:
This series applies a few cleanups to the test suite, and makes it easier to an installation on a Mac without having to build either `pgtap` or `pg_prove`.

(Although we most certainly don't support _running_ LoRA on a Mac, it's a somewhat common development platform.)

(Closes #87.)


oio_rest/oio_rest/utils/test_support.py
```diff
@@ -120,19 +120,20 @@ def is_template1(template):
...
         *funcs1,
         *templates1,
         *funcs2,
         *templates2,
-        *dblistdir('tests'),
     ]
     # </mess>
 
+    assert sorted(files) == sorted(set(files)), 'duplicates!!!!'
+
```
agger-magenta: If the files are listed from different directories, how could there be duplicates?

oio_rest/oio_rest/utils/test_support.py
```diff
@@ -42,6 +43,10 @@ def psql():
     return psql
 
 
+def list_db_sql(dirname):
+    return glob.glob(os.path.join(TOP_DIR, 'db', dirname, '*.sql'))
```
agger-magenta: I admit that this function is nicer than `dblistdir`, including the naming convention, but is it necessary to glob? Are there in fact non-SQL files to be excluded in these directories, so it's kind of redundant? (Just being curious, I don't think this is a showstopper).

oio_rest/requirements-test.txt
```diff
@@ -1,6 +1,7 @@
 -r requirements.txt
 Flask-Testing
 testing.postgresql
+tap.py
```
agger-magenta: :+1:

oio_rest/oio_rest/utils/test_support.py
danchr: > I admit that this function is nicer than `dblistdir`, including the naming convention, but is it necessary to glob? Are there in fact non-SQL files to be excluded in these directories, so it's kind of redundant? (Just being curious, I don't think this is a showstopper).

Yes, I ran into breakage when I had some `*.orig` files lying around from an earlier `hg revert`. Developers sometimes have weird cruft in their source trees 😉

oio_rest/oio_rest/utils/test_support.py
danchr: > If the files are listed from different directories, how could there be duplicates?

I don't think it could happen — but if it _did_ happen, it would definitely be an error. I added the assertion to prevent that in the future.

oio_rest/tests/test_sql.py
```diff
@@ -0,0 +1,46 @@
...
+
+            self.assertGreater(curs.rowcount, 0)
+
+            # tap.py doesn't support subtests yet, so strip the line
+            # see https://github.com/python-tap/tappy/issues/71
+            #
+            # please note that the tuple unpacking below is
+            # deliberate; we're iterating over over a cursor
+            # containing single-item rows
+            taptext = '\n'.join(line.strip() for (line,) in curs)
```
carlbordum: neat

oio_rest/tests/test_sql.py
```diff
@@ -0,0 +1,46 @@
...
+
+        for result in tap.parser.Parser().parse_text(taptext):
+            if result.category == 'test':
+                print(result)
+
+                with self.subTest(result.description):
+                    if result.skip:
+                        raise unittest.SkipTest()
```
carlbordum: If one, let's say the first test, is marked `skip`, does that mean no sql tests are run?

I'm asking because I would guess the raised exception cancels the function (`test_pgsql`) execution.

oio_rest/tests/test_sql.py
```diff
@@ -0,0 +1,46 @@
...
+                    curs.execute(fp.read())
+
+    def test_pgsql(self):
+        with self.db_cursor() as curs:
+            curs.execute("SELECT * FROM runtests ('test'::name)")
+
+            self.assertGreater(curs.rowcount, 0)
+
+            # tap.py doesn't support subtests yet, so strip the line
+            # see https://github.com/python-tap/tappy/issues/71
```
carlbordum: What's the custom for fixing stuff like this in Magenta?

oio_rest/tests/test_sql.py
danchr: That's the point of the [`self.subTest`](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) context manager — and in fact my main motivation for preferring `unittest` over `py.test`. The context manager suppresses any exceptions thrown within it, and allows evaluation to continue afterwards.

oio_rest/tests/test_sql.py
danchr: I'm not sure if we have any established policy in that regard, but I do so if it's not too much work. In this case, I'd have to look into and understand the TAP “procotol”, which seemed a bit much for something we really don't care all _that_ much about; we just want it to fail on errors.

carlbordum: There is an open PR that appears to just need one or two unittests to satisfy `coverage` requirements and be merged into master.

danchr: I _think_ I tried the PR, and it didn't help — but feel free to take a peak ;)

PR #100, opened 2018-12-17
Make mandatory OIO attributes NOT NULL in the database

carlbordum:
I adjusted the table generation and fixed failing tests. Please consider twice whether these changes are complete and if they have further repercussions as I am not 100% confident.


db/templates/tbls-specific.jinja.sql
```diff
@@ -114,11 +114,9 @@ ALTER TABLE {{oio_type}}_attr_{{attribut}}_id_seq
 
 CREATE TABLE {{oio_type}}_attr_{{attribut}} (
     id bigint NOT NULL DEFAULT nextval('{{oio_type}}_attr_{{attribut}}_id_seq'::regclass),
-    {%- for field in attribut_fields %} 
+    {% for field in attribut_fields %} 
        {{field}} {% if attributter_metadata is defined and attributter_metadata[attribut] is defined and attributter_metadata[attribut][field] is defined and attributter_metadata[attribut][field]['type'] is defined %}
-           {{attributter_metadata[attribut][field]['type']}}
-       {% else %}text{% endif %} null,
-    {%- endfor %} 
+           {{attributter_metadata[attribut][field]['type']}} {% else %}text{% endif %} {% if attributter_metadata is defined and attributter_metadata[attribut] is defined and attributter_metadata[attribut][field] is defined and attributter_metadata[attribut][field]['mandatory'] is defined and attributter_metadata[attribut][field]['mandatory'] %}not{% endif %} null,{% endfor %}
```
danchr: Could you insert a few newlines here? This is hard to read… FWIW Jinja2 has a [feature](http://jinja.pocoo.org/docs/2.10/templates/#whitespace-control) where you can strip whitespace around blocks with `{%-` and `-%}`.

(Also, the usual custom is to capitalise any SQL keywords, i.e. `NOT` and `NULL`, in this case.)

agger-magenta:
A quick comment is that there should always be a short description in the name of the feature branch, not just the ticket number.

db/templates/tbls-specific.jinja.sql
```diff
@@ -114,13 +114,17 @@ ALTER TABLE {{oio_type}}_attr_{{attribut}}_id_seq
...
     id bigint NOT NULL DEFAULT nextval('{{oio_type}}_attr_{{attribut}}_id_seq'::regclass),
-    {%- for field in attribut_fields %} 
-       {{field}} {% if attributter_metadata is defined and attributter_metadata[attribut] is defined and attributter_metadata[attribut][field] is defined and attributter_metadata[attribut][field]['type'] is defined %}
-           {{attributter_metadata[attribut][field]['type']}}
-       {% else %}text{% endif %} null,
-    {%- endfor %} 
-    virkning Virkning not null CHECK( (virkning).TimePeriod IS NOT NULL AND not isempty((virkning).TimePeriod) ),
-    {{oio_type}}_registrering_id bigint not null,
+    {%- for field in attribut_fields %}
+        {{ field }} {% if attributter_metadata is defined and attributter_metadata[attribut] is defined and attributter_metadata[attribut][field] is defined and attributter_metadata[attribut][field]['type'] is defined -%}
```
danchr: This chain of checks seems quite horrible. Can't we just generate a lookup dictionary that contains these?

agger-magenta:
I just tried testing this by checking out the branch and running `install.sh`. 

That, however, gave the following result:

```
ubuntu@moxtest:~/mox/interface_test$ psql mox -U mox
Password for user mox: 
psql (9.5.14)
Type "help" for help.

mox=> \d
No relations found.
mox=> 
```

Hence, all attempts at testing (of course) resulted in a `500 Internal Server Error`.

I'll try to install the DB manually, but this should really work out of the box.

carlbordum:
> A quick comment is that there should always be a short description in the name of the feature branch, not just the ticket number.

Jenkins fails with too long branch names, because it talks to postgres over a unix socket (that has max path length) and the unix socket name is generated from the branch name 😅

danchr:
> > A quick comment is that there should always be a short description in the name of the feature branch, not just the ticket number.
> 
> Jenkins fails with too long branch names, because it talks to postgres over a unix socket (that has max path length) and the unix socket name is generated from the branch name 😅

Yeah, that's a known bug 😞 Unfortunately, the runner has a tendency to leak database directories. I've been contemplating something like this:

```diff
# HG changeset patch
# User Dan Villiom Podlaski Christiansen <dan@magenta.dk>
# Date 1544196491 -3600
#      Fri Dec 07 16:28:11 2018 +0100
# Node ID 0697fde07efa64c71ddedae9d7cc3304fbb5643d
# Parent  4f85d700dcca77cbcc4bbab2e7da3c4a82c56321
test_support: use Jenkins temporary directory for db, or /tmp

This should both allow multiple concurrent runs, and decrease the
probability that we exceed the socket length limit.

diff --git a/oio_rest/oio_rest/utils/test_support.py b/oio_rest/oio_rest/utils/test_support.py
--- a/oio_rest/oio_rest/utils/test_support.py
+++ b/oio_rest/oio_rest/utils/test_support.py
@@ -14,6 +14,7 @@ import os
 import shutil
 import subprocess
 import sys
+import tempfile
 
 import click
 import mock
@@ -25,21 +26,29 @@ from .. import db_templating
 
 from .. import settings
 
+
 BASE_DIR = os.path.dirname(settings.__file__)
 TOP_DIR = os.path.dirname(BASE_DIR)
-DB_DIR = os.path.join(BASE_DIR, 'build', 'db')
 
 
 @functools.lru_cache()
 def psql():
-    os.makedirs(DB_DIR, exist_ok=True)
+    # the jenkins workspace directory
+    workspace_dir = os.getenv('WORKSPACE')
+
+    if workspace_dir:
+        db_dir = os.path.join(workspace_dir + '@tmp', 'db')
+    else:
+        db_dir = tempfile.mkdtemp(prefix='mox-db')
+
+    os.makedirs(db_dir, exist_ok=True)
 
     psql = testing.postgresql.Postgresql(
-        base_dir=DB_DIR,
+        base_dir=db_dir,
     )
 
     atexit.register(psql.stop)
-    atexit.register(lambda: shutil.rmtree(DB_DIR))
+    atexit.register(lambda: shutil.rmtree(db_dir))
 
     return psql
 
```

agger-magenta:
I don't think that the correct solution to the Jenkins/socket problem is to use shorter branch names. What do they do in other projects?

carlbordum:
What if we hash the branch name and pick the first n bytes of that?

db/templates/tbls-specific.jinja.sql
```diff
@@ -114,13 +114,17 @@ ALTER TABLE {{oio_type}}_attr_{{attribut}}_id_seq
...
+            {{ attributter_metadata[attribut][field]['type'] }}
+        {%- else -%}
+            text
+        {%- endif %} {% if attributter_metadata is defined and attributter_metadata[attribut] is defined and attributter_metadata[attribut][field] is defined and attributter_metadata[attribut][field]['mandatory'] is defined and attributter_metadata[attribut][field]['mandatory'] -%}
```
danchr:
```suggestion
        {%- endif -%}
        {% if attributter_metadata is defined and attributter_metadata[attribut] is defined and attributter_metadata[attribut][field] is defined and attributter_metadata[attribut][field]['mandatory'] is defined and attributter_metadata[attribut][field]['mandatory'] -%}
```

This is about as unreadable earlier line; we should break it up somehow. Perhaps a pre-calculated `dict`? (The above suggestion improves on that a bit, but doesn't really address the issue.)

oio_rest/oio_rest/oio_rest.py
```diff
@@ -345,6 +345,13 @@ def put_object(cls, uuid):
         input = cls.get_json()
         if not input:
             return jsonify({'uuid': None}), 400
+
+        # Validate JSON input
+        try:
+            validate.validate(input)
+        except jsonschema.exceptions.ValidationError as e:
+            return jsonify({'message': e.message}), 400
```
danchr: I suppose we do something similar elsewhere — describing a validation error should be in a dedicated, shared routine. Also, including the mere message probably isn't all that helpful — I think the exception object contains other fields might be helpful too, such as the `path` and `validator` field.

Also, do we have a test which exercises this code?

danchr:
> What if we hash the branch name and pick the first n bytes of that?

Jenkins does have some support for that, but unfortunately, its hashes are rather long 😒

db/templates/tbls-specific.jinja.sql
carlbordum: The jinja templates could be cleaner. A lot of these calculations are repeated multiple times and a lot of the if statements in for loops could be removed if we precalculated the right lists only once. This takes time and may be a worthy effort, but has nothing to do with this PR.

danchr: Fair enough — when writing this, I hadn't noticed that you merely edited these lines 🙂

oio_rest/oio_rest/oio_rest.py
danchr: Fair enough — I seem to recall having something like this elsewhere, but I seem to have misremembered.

PR #101, opened 2018-12-18
#25849 consolidate oio_rest

danchr:
This is a change I recently made — mostly on whim. Posting it here for discussion.

I've consolidated all the Python sources in the `oio_rest` directory into one module, as is customary. I also made sure that `setup.py` bundled all templates — including the database templates — as package data. The end result should be that the entire OIO REST server, including its database generation, is runnable when installed as a Python package.


carlbordum:
People want to experiment with docker. The general plan has been one container for the database and one for `oio_rest`, so I don't think this tighter coupling is appropriate right now.

That being said, I do like these changes and I think it is an option to be considered.

In terms of feedback on the actual code, maybe name `dbdata/` `db/` or at the very least `dbcode/`. I also liked the readme symlink you had created in the old db folder. This kills the installer. Remove the old `db/` folder completely and put `initdb.sh` somewhere in `installer/`. My 2¢.

danchr:
> People want to experiment with docker. The general plan has been one container for the database and one for `oio_rest`, so I don't think this tighter coupling is appropriate right now.

Well, the general idea is to move the application towards the regular Python application structure, where you have a database that's defined by the application. Both Django and Flask-SQLAlchemy follow this path. In other words, you'd have a Python package that's self-contained and able to create its own database.

agger-magenta:
> > People want to experiment with docker. The general plan has been one container for the database and one for `oio_rest`, so I don't think this tighter coupling is appropriate right now.
>  (...) In other words, you'd have a Python package that's self-contained and able to create its own database.

That also doesn't contradict having one container for the database. The DB container could run Postgres and nothing more.

danchr:
> > > People want to experiment with docker. The general plan has been one container for the database and one for `oio_rest`, so I don't think this tighter coupling is appropriate right now.
> > > (...) In other words, you'd have a Python package that's self-contained and able to create its own database.
> 
> That also doesn't contradict having one container for the database. The DB container could run Postgres and nothing more.

Indeed. But whatever _generates_ the database needs to have a Python VM. This PR doesn't change that — it merely changes what that Python VM does.

agger-magenta:
But the OIO Rest container could generate the database up against the postgres Docker container, I presume.

oio_rest/tests/test_sql.py
```diff
@@ -35,9 +35,8 @@ def setUp(self):
                 ),
             )
 
-            for dbfile in test_support.list_db_sql('tests'):
-                with open(dbfile) as fp:
-                    curs.execute(fp.read())
+            for dbfile in pathlib.Path(util.TESTS_DIR).glob("sql/*.sql"):
```
cmoesgaard: `list_db_sql` doesn't seem to be used anymore. Remove it.

oio_rest/setup.py
```diff
@@ -26,14 +35,27 @@
...
+
     entry_points={
         # -*- Entry points: -*-
         'console_scripts': [
-            'oio_api = oio_rest.app:app.run',
+            'mox = oio_rest.__main__:cli',
         ],
     },
+
+    dependency_links=[
```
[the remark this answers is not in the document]
danchr: I figured out a way to avoid using dependency links, but it requires a relatively new `pip` & `setuptools`…

oio_rest/oio_rest/__main__.py
```diff
@@ -0,0 +1,32 @@
...
+from . import db_templating
+
+
+@click.group(cls=flask.cli.FlaskGroup, create_app=lambda: app.app)
+def cli():
+    """Management script for OIO REST."""
+
+
+@cli.command()
+@click.option('-o', '--output', type=click.File('wt'), default='-')
```
carlbordum: can we consider something different here, ala `--output-sql` and maybe no single dash option?

oio_rest/oio_rest/__main__.py
```diff
@@ -0,0 +1,32 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import click
+import flask.cli
+
+from . import app
+from . import db_templating
+
+
+@click.group(cls=flask.cli.FlaskGroup, create_app=lambda: app.app)
```
carlbordum: `create_app=app.app`

oio_rest/oio_rest/db_templating.py
```diff
@@ -18,9 +18,9 @@
 
 from jinja2 import Environment, FileSystemLoader
 
-import settings
+from . import settings
 
-DB_DIR = Path(__file__).absolute().parent.parent.parent / "db"
+DB_DIR = Path(__file__).parent / "dbdata"
```
carlbordum: Please consider renaming this to `db` or `dbcode`. `dbdata` feels misleading to me.

oio_rest/setup.py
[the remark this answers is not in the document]
cmoesgaard: Neat! Looks like support for non-PyPI packages has become slightly less terrible. We should start doing this in our other projects as well...

oio_rest/run_tests.sh
```diff
@@ -31,7 +31,7 @@ export MOX_AMQP_PASS="guest"
 export MOX_AMQP_VHOST="/"
 
 # Execute tests
-$PYTHON -m pip install -e .
-$PYTHON -m pip install -r requirements-test.txt
+$PYTHON -m pip install --upgrade pip setuptools
```
carlbordum: Please add this to `install.sh` just before line 78. I suggest a new step called "Upgrade pip", but do it as you see fit.

oio_rest/oio_rest/__main__.py
danchr: I'd rather not; `-o/--output` is quite canonical when it comes to which argument should redirect the regular output somewhere else.

oio_rest/oio_rest/__main__.py
danchr: > create_app=app.app

Unfortunately, that doesn't work; the `create_app` argument must be a callable returning a `flask.Flask` instance, whereas `app.app` is such an instance. Making the change you propose results in this:

```python-traceback
Traceback (most recent call last):
  File "/private/tmp/venv/lib/python3.5/site-packages/flask/cli.py", line 529, in list_commands
    rv.update(info.load_app().cli.list_commands(ctx))
  File "/private/tmp/venv/lib/python3.5/site-packages/flask/cli.py", line 367, in load_app
    app = call_factory(self, self.create_app)
  File "/private/tmp/venv/lib/python3.5/site-packages/flask/cli.py", line 116, in call_factory
    return app_factory()
TypeError: __call__() missing 2 required positional arguments: 'environ' and 'start_response'
```

Instantiating the `Flask` instance _ought_ to be done in function…

oio_rest/oio_rest/__main__.py
carlbordum: But a rest interface with a cli that can ouput sql is not canonical hence my suggestion

danchr: > But a rest interface with a cli that can ouput sql is not canonical hence my suggestion

The name of the command is, literally, `sql`. You invoke it as:

```/path/to/venv/bin/mox sql```

The fact a command called `sql` outputs SQL shouldn't be all that surprising, should it? (If you have a suggestion for a better name, I'm all ears!)

Comment (carlbordum) [reply oio_rest/oio_rest/__main__.py]: Aha!

You're right.

I havent used `click` in years. Totally forgot that `sql` was a subcommand. Thought this implemented `mox --output`

File: oio_rest/oio_rest/db_templating.py (hunk 5 above)
Comment (danchr) [reply oio_rest/oio_rest/db_templating.py]: > Please consider renaming this to `db` or `dbcode`. `dbdata` feels misleading to me.

I consolidated everything into `oio_rest/db`, including relevant modules.

File: oio_rest/setup.py [hunk 7]
```diff
@@ -26,14 +43,28 @@
     license='Mozilla Public License Version 2.0',
     packages=find_packages(exclude=['ez_setup', 'examples', 'tests']),
     package_data={
-        '': ["*.txt", "*.xml", "*.pem"]
+        'oio_rest': [
+            'dbdata/*/*.sql',
```
carlbordum: !

carlbordum:
This is nothing blocking, but reconsider the directory names within `oio_rest/oio_rest/db`. There exists both `templates/` (which contains queries) and `sql/templates/` (which contains initializing database code).

PR #105, opened 2019-01-02
Remove installer

carlbordum:
This PR replaces the salt installer with a simpler `install.sh` script.

The purpose of this is produce a more familiar dev (no service, nginx, etc) environment and to get rid of the maintenance burden of the installer (costs us several hours some weeks).

@robertjensen you are asked to review, because this means we no longer install notification-service for the average dev. Any objections?

Please comment!


istherepie:
It has been on my TODO list for a while to remove the current installer and/or replace it with a simpler bash script. 

In my opinion the main goal of the installer is to perform a quick local install of the OIO rest application including the system dependencies/services with the purpose of the following: 

 - to use OIO Rest interface as working backend for developing integrations/addons
 - the development of the mox project itself. 

I cannot think of any scenarios where the other components (which have not been maintained for a while, such as "Mox Rest Frontend") would be installed locally - or even if these components are actually working and can be installed - The current installer does not include/install these components either. 

Either way the current/previous installer is NOT used in a production setting, as such this PR does not break any current functionality. 
In regards to the introduction of container technologies - this process will either way be handled by a build process / build server and does not rely on the "local" installer. 

I think the simpler bash installer improves on the install time and complexity added by the previous/current installer and improves on maintainability of the project by removing the hard coupling to the salt library.

agger-magenta:
The thing is that as developers we need to be able to quickly do a full-blown setup.

Maybe the thing to do would be to move the production installation into this repository, with sensible defaults that creates a working system, as the current installer and the one it superseded does/did, and have the pipeline salt installer provide the defaults, but use the installation files as provided here.

I don't see the gains of allowing only a development setup from here, when we often will need more than that.

agger-magenta:
So, yes

>Either way the current/previous installer is NOT used in a production setting ...

my conclusion is that it should be. The production installer should be in the repository, contain sensible defaults and be called by the pipeline with the production values. In that way we only have one installer, and it will be very fast to spin up a production-like machine in order to reproduce complicated bugs.

I agree we don't have to be able to install components that we don't maintain.

agger-magenta:
However, if we really do end up deciding to install everything with Docker components, all of our installation procedures will have to change anyway ... but then, we can decide _that_ when we know it. As is, I don't see any need for the changes introduced by this PR.

robertjensen:
For my part, I have no problems with removing amqp from the default install. No tools currently actually use amqp except for the proof-of-concept postgres-notification → amqp service. When (if) this service should be further developed, the developer should have no problems with installing the needed tools to do this.

danchr:
> So, yes
> 
> > Either way the current/previous installer is NOT used in a production setting ...
> 
> my conclusion is that it should be. The production installer should be in the repository, contain sensible defaults and be called by the pipeline with the production values. In that way we only have one installer, and it will be very fast to spin up a production-like machine in order to reproduce complicated bugs.

Ideally, we would have one installer — but considering that we have two that are _almost_ the same, I think @carlbordum's change is an improvement. I'd rather have something simple and obviously different that duplicates something else, instead of the current situation where we have two mostly equivalent — but subtly diverging — solutions.

install.sh
```diff
@@ -1,48 +1,174 @@
...
-# DOES NOT AUTOMATICALLY INSTALL THE FOLLOWING COMPONENTS:
-# Mox Rest Frontend, Mox Advis, Mox Elk log
+#!/usr/bin/env sh
+# Use install.sh to get a development environment up and running.
+# Don't use this in production.
+
+# This script does not automatically install the following components:
+#   Mox Rest Frontend, Mox Advis, Mox Elk log
+
+# Furthermore, it does not install rabbitmq or notification-service,
```
agger-magenta: But now it _does_ install `rabbitmq`, right?

agger-magenta:
If I make a new VM (Ubuntu Xenial, 2.5GB RAM), check out the code and run

`./install.sh --full`

everything seems to work fine, including
```
curl http://localhost/site-map
```

However, if I go to the `oio_rest` folder and run the tests, I get errors:

```
============= 16 failed, 362 passed, 9 skipped, 2 xfailed, 1 warnings in 23.98 seconds ==============
```

carlbordum:
@agger-magenta set localtime to Europe/Copenhagen

agger-magenta:
@carlbordum Argh, now I remember. We should definitely fix this! Having the test suite depend on the server's time setting doesn't really make sense (not shooting at you, I know this is an old bug).

PR #109, opened 2019-01-09
Run tests as if the computer was in Copenhagen

carlbordum:
The tests can now run *everywhere*!

(disclaimer: I only verified that they work in Copenhagen, Danmarkshavn (Grønland) and Mexico)

### DIFF
--- a/oio_rest/oio_rest/utils/test_support.py
+++ b/oio_rest/oio_rest/utils/test_support.py
@@ -76,6 +76,8 @@ def _initdb():
                 settings.DATABASE, settings.DB_PASSWORD,
             ))
 
+            curs.execute("ALTER DATABASE {} SET time zone 'Europe/Copenhagen'".format(settings.DATABASE))
+
     with psycopg2.connect(psql().url(
         database=settings.DATABASE,
     )) as conn:
--- a/oio_rest/oio_rest/utils/test_support.py
+++ b/oio_rest/oio_rest/utils/test_support.py
@@ -76,6 +76,11 @@ def _initdb():
                 settings.DATABASE, settings.DB_PASSWORD,
             ))
 
+            # The tests are written as if the computer has the local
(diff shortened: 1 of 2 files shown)


oio_rest/oio_rest/utils/test_support.py
```diff
@@ -76,6 +76,8 @@ def _initdb():
                 settings.DATABASE, settings.DB_PASSWORD,
             ))
 
+            curs.execute("ALTER DATABASE {} SET time zone 'Europe/Copenhagen'".format(settings.DATABASE))
```
danchr: We should add a comment prior to this line explaining why it's necessary, and it's not part of the templates. The obvious reason is that we do not, in fact, want to hard-code a time zone in production.

oio_rest/tests/test_integration_activity.py
```diff
@@ -478,7 +479,8 @@ def test_searching_with_limit_after_editing_bvn(self):
     def test_searching_temporal_order(self):
         objids = [str(uuid.UUID(int=i)) for i in range(3)]
 
-        no_time = datetime.datetime.now()
+        tz = dateutil.tz.gettz('Europe/Copenhagen')
+        no_time = datetime.datetime.now(tz).replace(tzinfo=None)
```
danchr: Move to a utility method?

oio_rest/tests/test_integration_activity.py
```diff
@@ -497,7 +499,7 @@ def test_searching_temporal_order(self):
 
         time.sleep(0.01)
 
-        exists_time = datetime.datetime.now()
+        exists_time = datetime.datetime.now(tz).replace(tzinfo=None)
```
danchr: Ditto?

oio_rest/tests/test_integration_activity.py
carlbordum: I don't believe in abstractions in the test suite :p

PR #112, opened 2019-01-16
Feature/25881 integrationsdata

carlbordum:
Integrationsdata added to all oio objects.

The changes to `Jenkinsfile` will be reverted before this is merged. This is because MO needs some changes too for this patch to work.

Closes [redmine#25881](https://redmine.magenta-aps.dk/issues/25881) and [redmine#25882](https://redmine.magenta-aps.dk/issues/25882)


cmoesgaard:
Could you also add `tilknyttedefunktioner` as a relation on `organisationfunktion`, that would allow us to completely do away with the `DB_STRUCTURE` override in MO

oio_rest/oio_common/db_structure.py
```diff
@@ -15,7 +15,7 @@
         "attributter": {
             "egenskaber": [
                 "brugervendtnoegle", "beskrivelse", "opbygning", "ophavsret",
-                "plan", "supplement", "retskilde"
+                "plan", "supplement", "retskilde", "integrationsdata",
```
danchr: How hard is it to make `integrationsdata` a JSON object? ISTR we ran into some issues with registration comparisons?

Jenkinsfile
```diff
@@ -36,7 +36,7 @@ pipeline {
     stage('Fetch MO') {
       steps {
         dir('mora') {
-          git url: 'https://github.com/magenta-aps/mora', branch: 'development'
+          git url: 'https://github.com/magenta-aps/mora', branch: 'feature/25881_remove_db_structure'
```
agger-magenta: This shouldn't be merged to development, right?

danchr: > This shouldn't be merged to development, right?

As Carl wrote:

> The changes to `Jenkinsfile` will be reverted before this is merged. This is because MO needs some changes too for this patch to work.

oio_rest/oio_common/db_structure.py
carlbordum: It's possible, but how would you search for something in the `integrationsdata` field on a query?

carlbordum:
> Could you also add `tilknyttedefunktioner` as a relation on `organisationfunktion`, that would allow us to completely do away with the `DB_STRUCTURE` override in MO

Done

Jenkinsfile
agger-magenta: OK

agger-magenta: If someone approves a pull request and it can be merged, they will sometimes merge it. You should state somewhere that this shouldn't be merged, e.g. by including DO NOT MERGE in the PR title.

oio_rest/oio_common/db_structure.py
danchr: > It's possible, but how would you search for something in the `integrationsdata` field on a query?

Considering that the use-case is indeed to stash JSON data in the string, I'm not sure having it JSON changes that. Anything searchable should be in a relation IMO.

db/tests/test_as_create_or_import_sag.sql
```diff
@@ -374,6 +374,7 @@ sagEgenskab := ROW (
 , true --principiel boolean,
 ,'sagsnummer_1' -- text,
 , 'titel_sag_1'-- text,
+, 'integrationsdata_1'-- text,
```
agger-magenta: It would be nicer with some space before the comment, e.g. we could say to spaces as convention before EOL comments.

carlbordum: I agree, but all the test files use a different (indentation, comma, whitespace, newline, comment) style, so I decided to just use the style used in the lines above my changes. I did so consistently

oio_rest/oio_common/db_structure.py
danchr: Having JSON in a string field is quite horrible; one of the advantages to modifying LoRA is that we can avoid that by either:

* Have the field be proper JSON in the database, or…
* Have the field be parsed and unparsed prior to stashing in the database, enforcing its JSON nature.

db/tests/test_as_create_or_import_sag.sql
agger-magenta: OK, we can clean that up another time if we become so inclined.

oio_rest/oio_common/db_structure.py
robertjensen: Vi er nødt til at kunne søge i feltet, ellers kan vi ikke finde et objekt ud fra integrationsdata uden at hive alle objekter ud.

agger-magenta: Det taler for at gøre det til et JSON-felt og få det indekseret.

agger-magenta: Hvis der er _behov_ for at søge ud fra integrationsdata, altså.

carlbordum: We /need/ to be able to search on integrationsdata and thats why its a text field. I know that a datatype such as `jsonb` would be better in terms of database searches and field integrity (although why do we care whether the json is malformed?), but how would you specify such a search in a query?

agger-magenta: We could devise a general search syntax for JSON fields.

One possiblity would be something like `?<JSONfieldname>::<fieldname>=value` and have the use of `::` only be valid if used with json fields.

robertjensen: That would certainly be a good solution, but it would also be a quite significant feature creep for this PR. Are we willing to put in the time required to implement this?

agger-magenta: If we don't do this I think it should just be a text field for now.

agger-magenta: If we had JSON fields, we could also have done away with the special handling of search words in Klasse (IIRC).

carlbordum: @agger-magenta sounds like a good solution and fun challenge. I'd love to implement it. Think its in scope?

Proposed syntax for nested fields?

agger-magenta: @carlbordum as for the scope I'd like input from @danchr and @cmoesgaard as well, but I'd rather do things properly than half-way.

agger-magenta: @carlbordum Could you make an estimate of what it would take to implement the JSON fields including indexing for search and search with a syntax as I sketched? Versus how we could search for the integration data with simple text search if we need it?

carlbordum: Sure

agger-magenta: Have you been looking at this estimate?

oio_rest/tests/test_db_helpers.py
```diff
@@ -456,7 +456,8 @@ def test_get_relation_names_default(self):
                 "tilknyttedeorganisationer",
                 "tilknyttedeitsystemer",
                 "tilknyttedeinteressefaellesskaber",
-                "tilknyttedepersoner"
+                "tilknyttedepersoner",
+                "tilknyttedefunktioner"
```
danchr: Although this is mostly a matter of taste, terminating commas have the advantage of not breaking `blame` whenever you add a new element:

```suggestion
                "tilknyttedefunktioner",
```

db/templates/as_list.jinja.sql
```diff
@@ -141,6 +141,7 @@ FROM
                                                        a.titel,
                                                        a.retskilde,
                                                        a.aendringsnotat,
+                                                       a.integrationsdata,
```
danchr: Ew. Not your fault, and not related to this PR, but do you know why the SQL code lists all the fields for `klasse` here? That seems quite horrible 😕

agger-magenta: That is a remnant of the old patch system. So this used to be part ofa patch for the generated files.

danchr: My question wasn't so much this exact code got there — but _why_ it's there in the first place, i.e. what does it do?

agger-magenta: This appears to be a special case handling of the search terms on klasse, check out this line:

https://github.com/magenta-aps/mox/blob/8db2371839496245056226796196844cb54a147a/db/templates/as_list.jinja.sql#L145

This would go away if we had JSON fields and used one of those to handle search terms. :slightly_smiling_face: 

But arguably, the handling of this special field could have been more elegant from the beginning.

oio_rest/tests/test_db_helpers.py
carlbordum: I agree, but I went with fewer (unrelated) diff lines and local consistency (the rest of the structure has many lists without terminating commas).

PR #115, opened 2019-01-22
26548: These files are no longer in use

istherepie:
MICRO CHANGE:
Leftovers from when the installer was replaced. 
Removing them before the upcoming release.


istherepie:
Gunicorn config is now contained in the systemd service file: 

`WorkingDirectory=$BASE_DIR
ExecStart=$BASE_DIR/python-env/bin/gunicorn \
    --bind 127.0.0.1:8080 \
    --log-syslog-prefix mox \
    --workers 4 \
    --access-logfile /var/log/mox/oio_access.log \
    --error-logfile /var/log/mox/oio_error.log \
    oio_rest.app:app
`

danchr:
> Gunicorn config is now contained in the systemd service file:
> 
> `WorkingDirectory=$BASE_DIR ExecStart=$BASE_DIR/python-env/bin/gunicorn \ --bind 127.0.0.1:8080 \ --log-syslog-prefix mox \ --workers 4 \ --access-logfile /var/log/mox/oio_access.log \ --error-logfile /var/log/mox/oio_error.log \ oio_rest.app:app`

Well, one advantage to having the Gunicorn configuration as a Python files is that it makes it easier to extract the application version, should we choose to do so.

istherepie:
Is it acceptable to keep the version number in the settings file instead?

danchr:
> Is it acceptable to keep the version number in the settings file instead?

Personally, I'd prefer to use the Python package version, and let the application somehow expose that using an API or command.

agger-magenta:
Like I just mentioned here, is the Python package version changed if you modify something only in the front end?

Anyway, I think:

* Any production installation will always run only on an official release.
* Any official release will always have a well-defined release number and release date.
* That information should be available  to sysadmins.

One solution would be to have a VERSION file like in LoRa and many other Magenta systems.

It could also be parsed from the top line of the `NEWS.rst` file.  It could also be in the settings, and that setting could even be parsed from one of the above options.

The only thing is that it's imperative that we can identify the installed version on any production site, and that it's relatively straightforward to do so.

istherepie:
In either way I think we should not keep the wsgi.py file just for displaying the version.

agger-magenta:
No, agreed since Salt doesn't use HTTP to communicate with MO anyway.

danchr:
> In either way I think we should not keep the wsgi.py file just for displaying the version.

Agreed — I mostly meant it as an aside; something to note when we go forward. This is good to go as-is 🙂

And continuing…

> Like I just mentioned here, is the Python package version changed if you modify something only in the front end?

Presumably, any deployment will be a _full_ deployment, so yes; it should be.

> Anyway, I think:
> 
> * Any production installation will always run only on an official release.
> * Any official release will always have a well-defined release number and release date.
> * That information should be available  to sysadmins.

Agreed.

> One solution would be to have a VERSION file like in LoRa and many other Magenta systems.
> 
> It could also be parsed from the top line of the `NEWS.rst` file. It could also be in the settings, and that setting could even be parsed from one of the above options.
> 
> The only thing is that it's imperative that we can identify the installed version on any production site, and that it's relatively straightforward to do so.

Currently, we extract the Python package version `package.json` — the description for the frontend sources. That's not terribly pretty. My point was rather to note that we _could_ use a `guniconfig.py` for that. Seems like we won't 🙂

Regardless, wouldn't a shell command for introspection be easiest on everyone?

istherepie:
A shell command seems like the sensible choice and will be relatively easy to maintain.

PR #119, opened 2019-01-28
#25849 simpler sql directory layout

danchr:
Just a simple reorganisation of SQL file layout to ensure ordering using numeric prefixes in filenames.


carlbordum:
I don't think we need to rush this into the next release. With faster release cycles it's not important.

Suggested change: add a readme or some documentation explaining the naming scheme.

I understand that this is a response to my snarky comment, that the project is starting to look like a java project (`oio_rest/oio_rest/db/sql/declarations/pre-funcs/_create_notify.sql`), but before I can approve this, I think we need to discuss whether we want to embed business logic in our file names

danchr:
> …but before I can approve this, I think we need to discuss whether we want to embed business logic in our file names

The general idea is that this conforms to the customary pattern seen in e.g. Django's migrations.

PR #121, opened 2019-01-29
Allow endpoints with and without trailing slashes.

carlbordum:
This has been the source of a bit too much confussion :)

Werkzeug interprets url rules without a trailing slash to exactly that, but
we're not that strict. The ``strict_slashes`` has been set to ``False`` for
all endpoints, which means that they 301 without the trailing slash to their
slashed counter part.

I took the liberty to reformat the calls to ``Flask.add_url_rule``. I hope I
didn't step on anyones toes.

### DIFF
--- a/oio_rest/oio_rest/app.py
+++ b/oio_rest/oio_rest/app.py
@@ -52,6 +52,7 @@ class RegexConverter(BaseConverter):
 
 
 app.url_map.converters['regex'] = RegexConverter
+app.url_map.strict_slashes = False
 
 app.teardown_request(db.close_connection)
 
--- a/oio_rest/oio_rest/app.py
+++ b/oio_rest/oio_rest/app.py
@@ -76,7 +76,7 @@ def root():
     return redirect(url_for('sitemap'), code=308)
 
 
-@app.route('/get-token', methods=['GET', 'POST'])
+@app.route('/get-token/', methods=['GET', 'POST'])
 def get_token():
     if request.method == 'GET':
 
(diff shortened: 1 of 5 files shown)


danchr:
Can't we do this at the higher level, i.e. the Flask app? Polluting each and every URL rule seems a bit ugly, and means that sooner or later this will turn up elsewhere.

Also, some of the paths — e.g. `site-map` and `get-token` — _aren't_ directories, so specifying the slash is wrong.

carlbordum:
@danchr to `werkzeug` a trailing slash does not mean directory, but rather that its the same resource with or without the slash. Without a trailing slash, `werkzeug` interprets it literally and will give `404` when a user tries to access a trailed url with a trailing slash. This is what has caused headache.

danchr:
> @danchr to `werkzeug` a trailing slash does not mean directory, but rather that its the same resource with or without the slash. Without a trailing slash, `werkzeug` interprets it literally and will give `404` when a user tries to access a trailed url with a trailing slash. This is what has caused headache.

A slash in a path is a directory separator, by definition. You're describing behaviour, I'm speaking of semantics. Yes, in a REST API the directory might merely be conceptual, but those concepts matter.

carlbordum:
FYI this doesn't have the behavior we agreed upon for `/site-map/` and `/get-token/`. I'm fine with this, though. We can explicitly set `strict_slashes` to `True` for those, if you _really_ want to behavior you described @danchr. /

danchr:
> FYI this doesn't have the behavior we agreed upon for `/site-map/` and `/get-token/`. I'm fine with this, though. We can explicitly set `strict_slashes` to `True` for those, if you _really_ want to behavior you described @danchr. /

Heh, but at least we aren't drawing _attention_ to them — I disliked that `/site-map/` appeared the canonical location. But now we officially just don't care :)

PR #124, opened 2019-01-31
tests: disable fsync in postgres, for speed; move socket to /tmp

danchr:
The former might cause a speedup, and who cares about resilience in
tests, anyway? The latter should address long branch names.


agger-magenta:
One question, though: Did you measure how much faster the tests run with fsync disabled?

PR #134, opened 2019-03-15
#27285 dummy test fixtures

danchr:
This PR exposes logic to easily load a dump using `psql`. This is primarily useful for fast loading within the MO testsuite.

https://redmine.magenta-aps.dk/issues/27285

### DIFF
--- a/oio_rest/oio_rest/utils/test_support.py
+++ b/oio_rest/oio_rest/utils/test_support.py
@@ -104,6 +104,17 @@ def psql():
     return psql
 
 
+def load_sql_fixture(fixture_path):
+    '''Empty the database and inject the given SQL fixture directly'''
+    with db.get_connection() as conn, conn.cursor() as curs:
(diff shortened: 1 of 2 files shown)


oio_rest/oio_rest/utils/test_support.py
```diff
@@ -104,6 +105,39 @@ def psql():
     return psql
 
 
+def load_sql_fixture(fixture_path):
+    '''Empty the database and inject the given SQL fixture directly.
+
+    We support optimised dumps thanks to invoking 'psql' rather than
```
agger-magenta: What is, more exactly and quantitatively, the advantage of using `psql` rather than `psycopg2`?

Additionally, where is this function supposed to be called?

agger-magenta: Maybe the answer to the last question is: From MO.

danchr: > Maybe the answer to the last question is: From MO.

Indeed. The advantage is that `psql` supports much faster dumps using `COPY`.

PR #140, opened 2019-04-05
Specify specific version of psycopg2 in requirements.

cmoesgaard:
We currently use PersistentConnectionPool which has been deprecated in version 2.8

A ticket has been opened to do the actual work of getting rid of the use of PersistentConnectionPool:
https://redmine.magenta-aps.dk/issues/28425

### DIFF
--- a/oio_rest/requirements.txt
+++ b/oio_rest/requirements.txt
@@ -1,6 +1,6 @@
 requests==2.21.0
 Flask==1.0.2
-psycopg2>=2.7
+psycopg2~=2.7.0
 pexpect==4.6.0
 python-dateutil==2.7.5
 pika==0.13.0


oio_rest/requirements.txt
```diff
@@ -1,6 +1,6 @@
 requests==2.21.0
 Flask==1.0.2
-psycopg2>=2.7
+psycopg2==2.7
```
danchr: 2.7 is massively out of date; either use 2.7.9 directly or use a compatibility version:
```suggestion
psycopg2~=2.7.0
```
Also; perhaps we should fix the root cause?

cmoesgaard: We should fix the root cause, yeah. This is just a fix until then.

PR #161, opened 2019-07-31
Validate patch requests with JSON schema

agger-magenta:


andreaskring:
@agger-magenta Looks good to me - I only have a single concern to raise: no new tests have been added to the `test_validate.py` that tests the `do_create=False` case. I think you should add a few tests that takes care of this use case, but it is of course up to you :-)

andreaskring:
(I have not run the test suite)

agger-magenta:
> I only have a single concern to raise: no new tests have been added to the test_validate.py that tests the do_create=False case. I think you should add a few tests that takes care of this use case, but it is of course up to you 

Yes, that makes sense. I'll add a couple of tests.

oio_rest/oio_rest/validate.py
```diff
@@ -419,26 +436,40 @@ def generate_json_schema(obj):
     return schema
 
 
+# Store already calculated schemas to save performance.
```
agger-magenta: This `StoredSchema` class is dead wood and should be removed.

cmoesgaard: I'll remove it.

PR #165, opened 2019-08-28
Do NOT log database password

carlbordum:
(and also use loggers, so we know where log msgs come from :-)

### DIFF
--- a/oio_rest/oio_rest/settings.py
+++ b/oio_rest/oio_rest/settings.py
@@ -32,17 +32,20 @@ import toml
 from oio_rest import app
 
 
+logger = logging.getLogger(__name__)
+
+
 def read_config(config_path):
     try:
         with open(config_path) as f:
             content = f.read()
     except FileNotFoundError as err:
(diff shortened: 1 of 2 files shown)


oio_rest/oio_rest/settings.py
```diff
@@ -32,17 +32,20 @@
...
+logger = logging.getLogger(__name__)
+
+
 def read_config(config_path):
     try:
         with open(config_path) as f:
             content = f.read()
     except FileNotFoundError as err:
-        logging.critical("%s: %r", err.strerror, err.filename)
+        logger.critical("%s: %r", err.strerror, err.filename)
```
danchr: Perhaps either pass `exc_info=True` or use `logging.exception()` to get the stack trace?

carlbordum: No. It logs "[INFO] Reading user config from {{filename}}", then critical FileNotFoundError with a filename again and immediately exits. That's clear.

PR #166, opened 2019-08-29
Remove db connection pool

carlbordum:


carlbordum:
Venter på Mathias' arbejde.

carlbordum:
Aha! We need to mention that somewhere. This is not a problem in docker, as it uses gunicorn and not the werkzeug dev server.

@danchr how do you start your werkzeug dev server? Reading the docs, it looks to me as it is single-threaded by default.

carlbordum:
Work continued in #170.

PR #168, opened 2019-09-04
Feature/28339 pg parametrisering

md-magenta:
[Redmine ticket #28339](https://redmine.magenta-aps.dk/issues/28339)

Her er sker tre store ændringer: 
- Docker-compose miljøet indeholder nu dependencies og korrekte database bruger rettigheder for at kunne køre testene.  Ticket [#28489](https://redmine.magenta-aps.dk/issues/28489)
- Alt test kode der forbinder til database går nu igennem `db/management.py`, bruger credentials fra settings istedet for at hardcodet strenge og er meget afkoblet fra databasen. `testing.postgresql`-pakken er fjernet.
- `oio_rest.test_support.TestCaseMixin` og `tests.util.TestCase` er erstattet af tre nye test klasser med hver deres funktionalitet: `tests.util.{ExtTestCase, DBTestCase, ExtDBTestCase}` der alle nedarver fra en base klasse.

Dette PR splitter MOs testsuite og Jenkins af og kan derfor ikke merges til `development` inden MO og gitlab-CI også er fikset. Derfor bliver det reviewet og merges det ind i `development-docker`-branchen. Der kommer sansynligvis flere PRs til `development-docker` inden MOs testsuite kører i docker og gitlab-ci. Derefter kan `development-docker` kan merges til `development`.


dev-environment/postgres-initdb.d/30-upgrade-to-superuser.sh
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+true "${DB_USER:?DB_USER is unset. Error!}"
+
+psql -v ON_ERROR_STOP=1 <<-EOSQL
+    ALTER ROLE $DB_USER WITH SUPERUSER;
+EOSQL
+
+# we can connect without password because ``trust`` authentication for Unix
+# sockets is enabled inside the container.
```
carlbordum: side note: is my "Do not replace this script with a raw .sql..." comments wrong in the other init scripts? I think what they point out is still true for shell scripts?

oio_rest/oio_rest/db/management.py
```diff
@@ -0,0 +1,225 @@
...
+
+    _dropdb(DBNAME)
+
+    if from_scratch:
+        _createdb(DBNAME)
+    else:
+        try:
+            # Only create template database if it does not exists.
+            _get_connection(DBNAME_INITIALIZED_TEMPLATE)
+        except psycopg2.OperationalError:
```
carlbordum: I am a bit worried this could lead to strange issues.
`_get_connection`  can fail for other reasons than `DBNAME_INITIALIZED_TEMPLATE` not existing. Besides, when does `template1` not exist?

oio_rest/oio_rest/db/management.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def testdb_reset(from_scratch=False):
+    """Reset the database specified in settings from the template. Requires the
+    template database to be created.
+
+    Requires CREATEDB and OWNER or SUPERUSER privileges.
+
+    """
+    logger.info("Reseting test database")
```
carlbordum: Reseting => Resetting

oio_rest/oio_rest/db/management.py
```diff
@@ -0,0 +1,225 @@
...
+            )
+
+
+def _dropdb(dbname):
+    """Deletes a pg database object.
+
+    Requires OWNER or SUPERUSER privileges.
+    """
+    close_pool()
+    logger.debug("Dropping database %s", dbname)
```
carlbordum: Jeg forstår ikke helt, hvornår en log besked har level `DEBUG` og level `INFO` i det her modul.

oio_rest/oio_rest/db/management.py
```diff
@@ -0,0 +1,225 @@
...
+            psycopg2.extensions.ISOLATION_LEVEL_AUTOCOMMIT
+        )
+        with conn.cursor() as curs:
+            curs.execute(
+                SQL("DROP DATABASE IF EXISTS {};").format(Identifier(dbname))
+            )
+
+def _createdb(dbname):
+    """Create a new database and initialize it with objects from templates.
+
```
carlbordum: Maybe, add: "Warning: drops database if it already exists." or a `force` flag although that is probably a YAGNI violation.

oio_rest/oio_rest/db/management.py
```diff
@@ -0,0 +1,225 @@
...
+        user=config["database"]["user"],
+        password=config["database"]["password"],
+        host=config["database"]["host"],
+        port=config["database"]["port"],
+    )
+
+
+def _cpdb(dbname_from, dbname_to):
+    """Copy a pg database object and add the attributes oio_rest expects.
+
```
carlbordum: Please add a paragraph in the docstring stating whether this copies data or only objects / structure.

oio_rest/oio_rest/db/management.py
```diff
@@ -0,0 +1,225 @@
...
+    Requires CREATEDB or SUPERUSER privileges.
+
+    """
+    close_pool()
```
carlbordum: Why? Is this in case other connections are open or lingering?

`close_pool()  # we cannot ... when other connections are open`

Although this does not handle the case, when there is an "external" connection, e.g. `psql` or `pgAdmin`. I don't think it should be "handled" when running the tests, but a proper error would be nice.

oio_rest/tests/util.py
```diff
@@ -33,7 +39,29 @@ def get_fixture(fixture_name, mode='rt', as_text=True):
             return fp.read()
 
 
-class TestCase(test_support.TestCaseMixin, flask_testing.TestCase):
+class _BaseTestCase(flask_testing.TestCase):
+    """Basic testcase without database support, but with various helper fuctions.
```
carlbordum: fuctions => functions

oio_rest/tests/util.py
```diff
@@ -260,3 +288,120 @@ def load_fixture(self, path, fixture_name, uuid=None):
...
+            yield
+
+
+class DBTestCase(_BaseTestCase):
+    """Testcase with database access. Will create a new database from a template
+    for each test. Requires the setting `[testing] enabled=True` during
+    database initialization to create the template.
+
+    This class does not support extending the database structure. For extension
+    tests use DBextTestCase.
```
carlbordum: DBextTestCase => ExtDBTestCase

doc/dev/test.rst
```diff
@@ -1,10 +1,28 @@
+.. _Testing:
```
carlbordum: Does it make sense to have a section in this file (or anywhere in the documentation) that mentions how to write tests? By this I mean, mentioning the test classes and their purposes and perhaps the distinction between test_integration_ and regular test_ files.

oio_rest/oio_rest/__main__.py
```diff
@@ -41,39 +47,26 @@ def initdb(wait):
...
     """
-    INIT_CHECK_SQL = (  # check that 'bruger' table exists. This is arbitrary.
-        "select relname"
-        "  from pg_catalog.pg_class"
-        " where relname = 'bruger';"
-    )
     SLEEPING_TIME = 0.25
 
     attempts = 1 if wait is None else int(wait // SLEEPING_TIME)
     for i in range(1, attempts + 1):
```
carlbordum: The loop body can be tidier and have better behavior in the case that the connection fails:
``` python
if check_connection():
    break
click.echo("Postgres is unavailable - attempt %s/%s" % (i, attempts))
if i == attempts:
    sys.exit(1)
time.sleep(SLEEPING_TIME)
```

doc/dev/test.rst
```diff
@@ -1,10 +1,28 @@
...
 
-OIO Rest has its own unit test suite::
+Most test dependencies are installed in the Docker image via
+:file:`oio_rest/requirements-test.txt`. The exception is `pgTAP
+<https://pgtap.org/>`_ which must be installed in the database.
+
+The tests use the database user credentials defined in :ref:`settings`. It
+requires the `CREATEDB
+<https://www.postgresql.org/docs/9.6/role-attributes.html>`_ privilege and
+`OWNER <https://www.postgresql.org/docs/9.6/sql-alterdatabase.html>`_ of the
```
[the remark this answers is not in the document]
md-magenta: No, mox only requires `GRANT ALL PRIVILEGES…` See [`10-init-db.sh`](https://github.com/magenta-aps/mox/blob/development/docker/postgres-initdb.d/10-init-db.sh#L11).

doc/user/installation.rst
```diff
@@ -15,11 +15,8 @@ below. Use the docker image or a Python package in a virtual environment.
       git clone https://github.com/magenta-aps/mox.git
       cd mox
       docker-compose up -d --build mox
-
-.. todo::
-
-   Document how to install test dependencies when it is formalized in #28489.
-
+      # run unittests with
+      docker-compose exec mox bash -c "cd oio_rest; python -m coverage run -m xmlrunner"
```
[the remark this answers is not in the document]
md-magenta: Yes. It is quite hidden. I think a mention in TL;DR is nice.

docker-compose.yml
```diff
@@ -20,11 +20,24 @@ services:
...
+      context: .
+      dockerfile: dev-environment/postgres.Dockerfile
     env_file:
       - dev-environment/db.env
+    # The following will mount the database data directory to RAM. It
+    # significantly speeds up the database at the cost of dataloss on shutdown.
+    # This is useful for running tests, but should never be used if you want to
+    # save the data.
+    # tmpfs:
+    #   - /var/lib/postgresql/data
```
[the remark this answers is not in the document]
md-magenta: If we want another docker-compose service for the testing database, we need to either monkeypatch hardcoded credentials to the new container when running tests or have a second instance of mox only for testing with other credentials and possibly `cmd: pytest`. It is doable, but the amount of services in docker-compose is already to long for my linking.

dev-environment/postgres-initdb.d/30-upgrade-to-superuser.sh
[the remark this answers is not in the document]
md-magenta: They are strictly correct, but misleading as the same applies to shell scripts. Both sql or shell script that exits early would still leave the database in a semi-initialised state and all scripts will be skipped on the next evocation.

I'll remove the comments from the other files.

oio_rest/oio_rest/db/management.py
md-magenta: The `template database` database refers to `DBNAME_INITIALIZED_TEMPLATE` not `template1`. What do you propose as an alternative?

Something like this?
```python
# […]
        def _check_database():
            with _get_connection(DBNAME_SYS_TEMPLATE) as conn:

                with conn.cursor() as curs:
                    curs.execute(
                        "select datname from pg_catalog.pg_database where datname=%s",
                        [DBNAME_INITIALIZED_TEMPLATE],
                    )
                    if curs.fetchone():
                        return True
            return False

        if not _check_database():
            _createdb(DBNAME_INITIALIZED_TEMPLATE)
# […]
```

doc/dev/test.rst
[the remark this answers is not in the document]
carlbordum: You are correct. It got mixed up in my head when we talked about doing #167 from Python. Reminder that the documentation states that we are the database owner (even when not running tests).

oio_rest/oio_rest/db/management.py
md-magenta: Yes. If you run the services with `docker-compose up mox` and then run `docker-compose exec mox bash -c "cd oio_rest; python -m coverage run -m xmlrunner"`, the `oio_rest` is still running and the connections are left open. Do you propose to remove this and only support the use of `docker-compose run` without the service running?

oio_rest/oio_rest/db/management.py
md-magenta: Offentlige metoder har `INFO`. Private metoder har `DEBUG`.

PR #171, opened 2019-09-19
Feature/31797 testdb api

md-magenta:
Tilføjer et REST API til at resette databasen.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,12 @@ services:
     volumes:
       - .:/code
       - ./dev-environment/dev-settings.toml:/user-settings.toml
-    command: ["gunicorn", "--reload", "-b", "0.0.0.0:8080", "oio_rest:app"]
+    command: [
+    "gunicorn",
+    "--config", "docker/gunicorn-settings.py",
+    "--reload", # Reload on file changes
+    "--workers", "1", # Single threaded in dev
+    "oio_rest:app"]
     ports:
       - "8080:8080"
     env_file:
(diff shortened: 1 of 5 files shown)


oio_rest/oio_rest/default-settings.toml
```diff
@@ -98,3 +98,10 @@ path = ""
 [log]
 activity_log_path = "/var/log/mox/activity.log"
 trace_log_path = "/var/log/mox/trace.log"
+
+
+[testing_api]
+# If enabled, expose /testdb/* endpoint for setup, reset and teardown of a test
+# database. Useful for integration test from other componenets such as MO. Does
+# not work when running multi-threaded.
```
carlbordum: test => tests
componenets => components

Why does it not work when running multithreaded?

docker-compose.yml
```diff
@@ -11,7 +11,12 @@ services:
     volumes:
       - .:/code
       - ./dev-environment/dev-settings.toml:/user-settings.toml
-    command: ["gunicorn", "--reload", "-b", "0.0.0.0:8080", "oio_rest:app"]
+    command: [
+    "gunicorn",
+    "--config", "docker/gunicorn-settings.py",
+    "--reload", # Reload on file changes
+    "--workers", "1", # Single threaded in dev
```
carlbordum: Why?

oio_rest/oio_rest/default-settings.toml
md-magenta: Se other reply.

docker-compose.yml
md-magenta: When MO hits several ``/testing/db-*` endpoints in a row, it alternates between the multiple workers and leaves idle db connections in the old threads.

PR #172, opened 2019-09-24
add /version endpoint

carlbordum:


oio_rest/oio_rest/views.py
```diff
@@ -108,6 +108,22 @@ def sitemap():
...
 
+_version = None
+
+
+def _get_version():
+    if _version is None:
+        global _version
+        with open(settings.config["version_file_path"], "r") as f:
+            _version = f.read()
+    return _version
```
cmoesgaard: I think we should return the version in a similar format to how we do it for OS2mo (https://github.com/OS2mo/os2mo/pull/414).

e.g.
```
{
    "lora_version": "1.3.0"
}
```

